Issue #13, opened 2015-02-18
Need a test to handle errors in callbacks


mattheworiordan:
We should log the errors and continue and not let a callback stop the library

mattheworiordan:
Done, https://github.com/ably/ably-ruby/pull/16

PR #16, opened 2015-03-06
Catch exceptions in callbacks

mattheworiordan:
@paddybyers and @kouno, I now ensure that any callbacks used anywhere externally from the library (public interfaces that the library provides) can no longer raise exceptions that will stop the client library from proceeding.  Instead errors are caught and logged to the logger as an error, and processing continues.

As you can see, it involved a little more code than I had hoped, but it least it has good test coverage! ;)

This solves https://github.com/ably/ably-ruby/issues/13

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,0 +1,1 @@
+--color --format documentation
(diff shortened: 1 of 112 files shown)


mattheworiordan:
Merged with https://github.com/ably/ably-ruby/commit/bcafc09480c3b7afb966b714d10eff16c5d79e1e

Issue #11, opened 2015-02-10
Add test coverage for receiving messages more than once i.e. historical messages resent somehow on reconnect


mattheworiordan:
The client library should be smart enough to know that if a message is received for a serial number it has already received it is simply discarded and an error is logged.

mattheworiordan:
This is tested, https://github.com/ably/ably-ruby/blob/bf0aff4942c905a6f76a242388bdae8f31bef56b/spec/acceptance/realtime/message_spec.rb#L279-L309

PR #17, opened 2015-03-06
Ensure messages received once

mattheworiordan:
@paddybyers and @kouno, I handle protocol errors better in the library now, see https://github.com/ably/ably-ruby/issues/11.  @paddybyers, is my behaviour in line with yours if I receive a protocolmessage for an already received connectionSerial?


mattheworiordan:
Merged with https://github.com/ably/ably-ruby/commit/bcafc09480c3b7afb966b714d10eff16c5d79e1e

PR #22, opened 2015-03-23
History since attach

mattheworiordan:
@kouno and @paddybyers following our chat, here is my proposal to support history requests on real-time channels that only give you messages published before the channel attached.

Please tell me what you think as we should probably add this or something like this to the other real-time client libraries.

### DIFF
--- a/lib/ably/realtime/channel.rb
+++ b/lib/ably/realtime/channel.rb
@@ -71,6 +71,11 @@ module Ably
       # @api private
       attr_reader :manager
 
+      # Serial number assigned to this channel when it was attached
+      # @return [Integer]
+      # @api private
+      attr_reader :attached_serial
+
       # Initialize a new Channel object
       #
       # @param  client [Ably::Rest::Client]
@@ -235,7 +240,7 @@ module Ably
       end
 
       private
-      attr_reader :queue, :attached_serial
+      attr_reader :queue
 
       def setup_event_handlers
         __incoming_msgbus__.subscribe(:message) do |message|
--- a/lib/ably/realtime/channel.rb
+++ b/lib/ably/realtime/channel.rb
@@ -23,7 +23,7 @@ module Ably
     #   Channel::STATE.Detached
     #   Channel::STATE.Failed
     #
-    # Channels emit errors - use `on(:error)` to subscribe to errors
+    # Channels emit errors - use +on(:error)+ to subscribe to errors
     #
     # @!attribute [r] state
     #   @return {Ably::Realtime::Connection::STATE} channel state
@@ -77,7 +77,7 @@ module Ably
       # @param  name [String] The name of the channel
       # @param  channel_options [Hash]     Channel options, currently reserved for Encryption options
       # @option channel_options [Boolean]  :encrypted       setting this to true for this channel will encrypt & decrypt all messages automatically
-      # @option channel_options [Hash]     :cipher_params   A hash of options to configure the encryption. *:key* is required, all other options are optional.  See {Ably::Util::Crypto#initialize} for a list of `cipher_params` options
+      # @option channel_options [Hash]     :cipher_params   A hash of options to configure the encryption. *:key* is required, all other options are optional.  See {Ably::Util::Crypto#initialize} for a list of +cipher_params+ options
       #
       def initialize(client, name, channel_options = {})
         ensure_utf_8 :name, name
@@ -181,13 +181,25 @@ module Ably
 
       # Return the message history of the channel.
       #
+      # Once attached to a channel, you can retrieve messages published on the channel before the
+      # channel was attached with the option <tt>:end => :before_attach</tt>.  This is very useful for
+      # developers who wish to display both new realtime messages as well as historical messages with
+      # the guarantee that no messages have been missed.
+      #
       # @param (see Ably::Rest::Channel#history)
       # @option options (see Ably::Rest::Channel#history)
+      # @option options [Integer,Time,Symbol]   :end        Time or millisecond since epoch, or +:before_attach+ for all messages published before this channel has been attached.
       #
       # @yield [Ably::Models::PaginatedResource<Ably::Models::Message>] First {Ably::Models::PaginatedResource page} of {Ably::Models::Message} objects accessible with {Ably::Models::PaginatedResource#items #items}.
       #
       # @return [Ably::Util::SafeDeferrable]
       def history(options = {}, &callback)
+        if options[:end] == :before_attach
+          raise ArgumentError, 'option :before_attach cannot be specified if the channel is not attached' unless attached?
+          options[:from_serial] = attached_serial
+          options.delete :end
+        end
+
         async_wrap(callback) do
           rest_channel.history(options.merge(async_blocking_operations: true))
         end
@@ -212,6 +224,11 @@ module Ably
         @error_reason = nil
       end
 
+      # @api private
+      def set_attached_serial(serial)
+        @attached_serial = serial
+      end
+
       # Used by {Ably::Modules::StateEmitter} to debug state changes
       # @api private
       def logger
@@ -219,7 +236,7 @@ module Ably
       end
 
       private
-      attr_reader :queue
+      attr_reader :queue, :attached_serial
 
       def setup_event_handlers
         __incoming_msgbus__.subscribe(:message) do |message|
--- a/lib/ably/realtime/channel.rb
+++ b/lib/ably/realtime/channel.rb
@@ -179,25 +179,24 @@ module Ably
         @presence
       end
 
-      # Return the message history of the channel.
+      # Return the message history of the channel
       #
       # Once attached to a channel, you can retrieve messages published on the channel before the
-      # channel was attached with the option <tt>:end => :before_attach</tt>.  This is very useful for
-      # developers who wish to display both new realtime messages as well as historical messages with
+      # channel was attached with the option <tt>until_attach: true</tt>.  This is very useful for
+      # developers who wish to subscribe to new realtime messages yet also display historical messages with
       # the guarantee that no messages have been missed.
       #
       # @param (see Ably::Rest::Channel#history)
       # @option options (see Ably::Rest::Channel#history)
-      # @option options [Integer,Time,Symbol]   :end        Time or millisecond since epoch, or +:before_attach+ for all messages published before this channel has been attached.
+      # @option options [Boolean]  :until_attach  When true, request for history will be limited only to messages published before this channel was attached. Channel must be attached.
       #
       # @yield [Ably::Models::PaginatedResource<Ably::Models::Message>] First {Ably::Models::PaginatedResource page} of {Ably::Models::Message} objects accessible with {Ably::Models::PaginatedResource#items #items}.
       #
       # @return [Ably::Util::SafeDeferrable]
       def history(options = {}, &callback)
-        if options[:end] == :before_attach
-          raise ArgumentError, 'option :before_attach cannot be specified if the channel is not attached' unless attached?
+        if options.delete(:until_attach)
+          raise ArgumentError, 'option :until_attach cannot be specified if the channel is not attached' unless attached?
           options[:from_serial] = attached_serial
-          options.delete :end
         end
 
         async_wrap(callback) do
--- a/SPEC.md
+++ b/SPEC.md
@@ -1,22 +1,27 @@
-# Ably Real-time & REST Client Library 0.7.5 Specification
+# Ably Realtime & REST Client Library 0.7.5 Specification
 
 ### Ably::Realtime::Channel#history
 _(see [spec/acceptance/realtime/channel_history_spec.rb](./spec/acceptance/realtime/channel_history_spec.rb))_
   * using JSON and MsgPack protocol
-    * [returns a SafeDeferrable that catches exceptions in callbacks and logs them](./spec/acceptance/realtime/channel_history_spec.rb#L20)
+    * [returns a SafeDeferrable that catches exceptions in callbacks and logs them](./spec/acceptance/realtime/channel_history_spec.rb#L21)
     * with a single client publishing and receiving
-      * [retrieves real-time history](./spec/acceptance/realtime/channel_history_spec.rb#L33)
+      * [retrieves realtime history](./spec/acceptance/realtime/channel_history_spec.rb#L34)
     * with two clients publishing messages on the same channel
-      * [retrieves real-time history on both channels](./spec/acceptance/realtime/channel_history_spec.rb#L45)
+      * [retrieves realtime history on both channels](./spec/acceptance/realtime/channel_history_spec.rb#L46)
     * with lots of messages published with a single client and channel
       * as one ProtocolMessage
-        * [retrieves history forwards with pagination through :limit option](./spec/acceptance/realtime/channel_history_spec.rb#L87)
-        * [retrieves history backwards with pagination through :limit option](./spec/acceptance/realtime/channel_history_spec.rb#L96)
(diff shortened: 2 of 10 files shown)


mattheworiordan:
@kouno and @paddybyers, do you have a view on this as this does have an effect on our other client library APIs.  If not, I will merge however the Ruby spec is what client lib developers reference.

mattheworiordan:
@paddybyers and @kuono? Your feedback on this would be good as it impacts on the documentation and client libraries currently being developed

mattheworiordan:
Mean to be @kouno

paddybyers:
> Your feedback on this would be good

I still don't like the way that it confuses `start`/`end` - which are meant to represent a time or date - with a specific message serial. But I also don't like the idea of introducing another param (eg `startSerial`) just in order to be able to give it this symbolic value for this use-case.

mattheworiordan:
Well we need to make a decision one way or another.  I personally think it's a good fit because your'e saying "Give me everything up to Time / epoch / time of channel attached", but agree in typed languages this could be a bit tricker.  I'm open to suggestions, but would like to close this PR so that we can include this in our spec as I think it's going to be a frequently needed feature.

lib/ably/realtime/channel.rb
```diff
@@ -179,15 +179,28 @@ def presence
...
       #
       # @param (see Ably::Rest::Channel#history)
       # @option options (see Ably::Rest::Channel#history)
+      # @option options [Integer,Time,Symbol]   :end        Time or millisecond since epoch, or +:before_attach+ for all messages published before this channel has been attached.
       #
       # @yield [Ably::Models::PaginatedResource<Ably::Models::Message>] An Array of {Ably::Models::Message} objects that supports paging (#next_page, #first_page)
       #
       # @return [Ably::Util::SafeDeferrable]
       def history(options = {}, &callback)
+        if options[:end] == :before_attach
```
kouno: If you are going this way, shouldn't we also provide a `start: :after_attach` option.

lib/ably/realtime/channel.rb
```diff
@@ -179,15 +179,28 @@ def presence
...
+          raise ArgumentError, 'option :before_attach cannot be specified if the channel is not attached' unless attached?
```
kouno: I'm not sure you should raise an exception here. It sounds to me that it should just attach and do the query at the moment it receives the attach confirmation.

lib/ably/realtime/channel.rb
mattheworiordan: @kouno, well perhaps, @paddybyers is that supported?

lib/ably/realtime/channel.rb
mattheworiordan: @kouno I did think about this, but it seems a bit weird that someone is asking for history prior to the channel attaching when the channel is not even attached. I think it's reasonable to assume that a developer should not request history prior to attach when not attached.  Also, if they requested this and the state of the channel was not attached, do we wait indefinitely before raising this error?

kouno: Hum, I'm okay to say that a developer should know about a connection and the connection lifecycle, but raising an exception is painful IMO.

We should just log it and warn them that they are using this option incorrectly. Not potentially break their whole app.

paddybyers: > it seems a bit weird that someone is asking for history prior to the channel attaching when the channel is not even attached

The whole point is that you can subscribe, which means that messages from the moment of subscription are sent to your listener, and then this functionality allows you to get exactly the messages that occurred before you attached. So it is only meaningful if there is a call to `subscribe()` prior to the call to `history()` using this param. Since the `subscribe()` implicitly calls `attach()` there is no sense in the `attach()` also being called later.

> We should just log it and warn them that they are using this option incorrectly. Not potentially break their whole app.

An exception isn't an inconvenience to a programmer; it's a way to tell him that the program he wrote cannot meaningfully continue. So if an exception is called for, it helps him.

kouno: > An exception isn't an inconvenience to a programmer; it's a way to tell him that the program he wrote cannot meaningfully continue. So if an exception is called for, it helps him.

Raising an exception here is an inconvenience because the program _can_ continue. `before_attach` just means that we want to retrieve any message before the attach event occurred. In our case, we can very well do that since we are not attached yet.

mattheworiordan: > Raising an exception here is an inconvenience because the program can continue

The program can continue, but the history request may continue with unpredictable results for a developer.  Will it timeout if the attach does not happen, if the connection drops should this fail or wait for it to reconnect and then attach, if whilst in the `attaching` state someone tells it to detach should it fail, how long should it wait for the channel to attach etc.  There are way too many permutations that we need to think about make assumptions for, whereas if we simply raise an exception then we leave that decision to the developer, which is certainly my preference.

kouno: I got your point, but it sounds weird to say that we can do a `history` call with an async call to `attach` behind the scene without the developer worrying, and at the same time we can't handle an options that would require that implicit `attach`.

paddybyers: I do not agree that there should be an implicit `attach()`

kouno: > I do not agree that there should be an implicit attach()

And suddenly it all makes sense. :smile:

mattheworiordan: There is no implicit `attach` because it's perfectly valid to ask for history on a channel you are not attached to and have no intention of attaching to.  Added to this, asking for "Give me everything before I attached" when you're not attached is simply wrong.  You shouldn't be asking for the history of a future event.

mattheworiordan: @kouno read the code!

kouno: > There is no implicit attach because it's perfectly valid to ask for history on a channel you are not attached to and have no intention of attaching to.

No implicit attach? Ah, I got it, I thought `attach` was for the connection, but it's for the channel :/. Sorry, wrong vocabulary.

> @kouno read the code!

Hey, listen... I read code, but I can't magically get all of it in my head overnight. You don't have to be like that.

mattheworiordan: We've had 12 posts about this point and you've been adamant about how this functionality should work, yet 4 lines beneath this post you can see that no attach is implicitly made or required.  It wasn't meant to come across as negative, it was meant to say please read the code before you have a strong opinion on how the code should work. That's not unreasonable.

kouno: I'm sorry but that's not true. Your async wrapper does a websocket connect, and `subscribe`/`publish` are doing an implicit `attach`. I don't think I was far off thinking that `history` required the same.

I would have to compare your `subscribe`/`publish` implementation, and see what is inside of the state machine to know what is going on. So no. It's not that simple.

mattheworiordan: `async` does not such thing, it just wraps a synchronous call in an asynchronous `Deferrable` and runs this in a thread.    See the documentation at http://www.rubydoc.info/gems/ably/0.7.5/Ably/Modules/AsyncWrapper.  This request has nothing to do with websocket connections, `subscribe` or `publish`.

kouno: Ok fine. I'm tired and seem to have misread the whole history part. I'll leave you two decide on this.

mattheworiordan:
Agreed to try `until_attach: true` and see if that feels right.

mattheworiordan:
@paddybyers and @kouno here is the API change we agreed, see 3229002.  I also added support on the presence history in 02d87c9.

I'd like to merge ASAP, can you tell me if you're happy and I'll merge in so that the spec is updated as well.

mattheworiordan:
@paddybyers and @kouno, still no feedback on this.  Can I merge and make this our spec for our client libraries now?

lib/ably/realtime/channel.rb
```diff
@@ -179,15 +184,26 @@ def presence
         @presence
       end
 
-      # Return the message history of the channel.
+      # Return the message history of the channel
+      #
+      # Once attached to a channel, you can retrieve messages published on the channel before the
```
kouno: I think the documentation should mention the following points:
1- `history` is only a wrapper for the rest channel history method. It therefore takes the same `options` as the REST `history` call. In addition since it is evented, you can pass a block as a callback once the REST call is complete.
2- You can use the `:until_attach` option if you need to display history with a guarantee that no message will be missed.
3- To use `until_attach` you need to make sure that the channel is attached before calling `history`.
4- If the channel is not attached, this option will throw an exception.

kouno:
Looks good to me. The documentation is my only last point, but I'm not sure how verbose it needs to be.

lib/ably/realtime/channel.rb
mattheworiordan: I am a bit confused by this comment.

> 1- history is only a wrapper for the rest channel history method. It therefore takes the same options as the REST history call. 

If you look at the [slightly out of date docs](http://www.rubydoc.info/gems/ably/0.7.5/Ably/Realtime/Channel#history-instance_method) it does provide documentation for all the options.

>  In addition since it is evented, you can pass a block as a callback once the REST call is complete.

Every method in the Realtime library is a deferrable and yields a block as well, this is again quite clear in http://www.rubydoc.info/gems/ably/0.7.5/Ably/Realtime/Channel#history-instance_method

>  You can use the :until_attach option if you need to display history with a guarantee that no message will be missed.

The comments say "Once attached to a channel, you can retrieve messages published on the channel before the channel was attached with the option <tt>until_attach: true</tt>.  This is very useful for developers who wish to subscribe to new realtime messages yet also display historical messages with the guarantee that no messages have been missed."

> To use until_attach you need to make sure that the channel is attached before calling history.

It says that in the option description: ":until_attach  When true, request for history will be limited only to messages published before this channel was attached. Channel must be attached."

> 3- If the channel is not attached, this option will throw an exception.

As above

kouno: > The comments say "Once attached to a channel, you can retrieve messages published on the channel before the channel was attached with the option until_attach: true. This is very useful for developers who wish to subscribe to new realtime messages yet also display historical messages with the guarantee that no messages have been missed."

I know. I find it lacking. "Once attached to a channel" doesn't tell the user that he __needs__ to attach this channel to use this feature.

The documentation tells that it is "uncommon" to call `attach` since it is implicit with `publish` and `subscribe`: https://github.com/ably/ably-ruby/blob/4425f4e3fba3df9a2ac7cdd23c0fe5cd6945222f/lib/ably/realtime/channel.rb#L154-L155.

`publish` and `subscribe` do not mention `attach` at all: https://github.com/ably/ably-ruby/blob/4425f4e3fba3df9a2ac7cdd23c0fe5cd6945222f/lib/ably/realtime/channel.rb#L103, https://github.com/ably/ably-ruby/blob/4425f4e3fba3df9a2ac7cdd23c0fe5cd6945222f/lib/ably/realtime/channel.rb#L130.

Even `presence` does an implicit `attach` https://github.com/ably/ably-ruby/blob/4425f4e3fba3df9a2ac7cdd23c0fe5cd6945222f/lib/ably/realtime/channel.rb#L183.

Actually `history` is the only one not doing an implicit `attach` on the public method of this class (apart from `unsubscribe` for obvious reason I guess). And it is on the other hand the only one mentioning any relation to `attach` on a specific option.

I'm also wondering what happens in `unsubscribe` and if it auto-detach from the channel too (which would sound reasonable). https://github.com/ably/ably-ruby/blob/4425f4e3fba3df9a2ac7cdd23c0fe5cd6945222f/lib/ably/realtime/channel.rb#L142-L151

kouno: "Channel must be attached." is also lacking in the fact that it is not telling what happens if you don't do it.

mattheworiordan: Ok, well I'd welcome a PR then...

mattheworiordan: I am not sure I agree on your second comment.  Imagine a method with an argument definition such as:
```ruby
# id is required 
def get_thing(id)
end
```

Do we really need to tell people what will happen if they don't follow the rules of a method definition?

mattheworiordan: The alternative would be:
```ruby
# id is required 
#  will raise a ArgumentError if nil
#  will raise a RecordNotFoundError if ID is missing
def get_thing(id)
end
```

That seems unnecessary IMHO

kouno: The problem with your example is that Ruby will blow up if you don't pass `id`. On the other hand `option` is a hash. And you could technically add anything you want in it.

```ruby
# works.
channel.history(something: 'hello')

# blows up
channel.history(until_attach: true)
```

mattheworiordan: It's an arguments hash, so no different.

kouno: It is different in that what you describe as an argument error is going to fail only on certain occasion.

```ruby
# this fails all the time because you have not specified `id`
get_thing

# this may fail or not depending on the channel state
channel.history(until_attach: true)

# this never fails because it is just ignored
channel.history(something: 'hello')
```

kouno: Maybe it would make sense to throw a custom exception in this case since it is dependent of the Channel state and that, in fact, the option is valid?

mattheworiordan: Sure custom exception is fine. PR?

paddybyers: > Actually history is the only one not doing an implicit attach

I said this already; I think you're missing the point.

For this functionality to be relevant, you have to have called `subscribe()` before calling `history()`. There is no situation where calling `attach()` implicitly as part of the `history()` call makes any sense.

kouno: > > Actually history is the only one not doing an implicit attach

> I said this already; I think you're missing the point.
> For this functionality to be relevant, you have to have called subscribe() before calling history(). There is no situation where calling attach() implicitly as part of the history() call makes any sense.


I listened and now I'm pointing out something else. I did not say it should do an implicit attach, I'm just pointing out that it is the only one which does not do an implicit attach.

Which calls for explanation in our documentation I think, especially since we add this `until_attach` option which is Channel connection aware.

All I'm saying is that the wording could be better.

mattheworiordan: >  I did not say it should do an implicit attach, I'm just pointing out that it is the only one which does not do an implicit attach.

`publish` and `unsubscribe` do not implicitly attach.

kouno: `publish` calls queue_message https://github.com/ably/ably-ruby/blob/6a5f9607f18519ae0f58927d11a187371cda4091/lib/ably/realtime/channel.rb#L121 which in turn calls `attach` if it's not already https://github.com/ably/ably-ruby/blob/6a5f9607f18519ae0f58927d11a187371cda4091/lib/ably/realtime/channel.rb#L240-L244.

As for unsubscribe, my question was "does it implicitly detach too?". It's calling `super` but I believe if you did not specify anything in the method it should not, right?

mattheworiordan: Ah, true, `publish` does implicitly attach.

See https://github.com/ably/ably-ruby/commit/c5292f8738d166b96e911e2eced794b1dbda7270 for a fix to the docs.

paddybyers: > As for unsubscribe, my question was "does it implicitly detach too?"

No, because you're just removing a listener, and there might of course be multiple other listeners still registered. And no, we don't count the listeners because you might still want to `publish()`.

kouno: So once attached, the only way to detach is to call `detach`... Or have a connection failure I suppose?

Anyway, makes sense. Thank you.

PR #29, opened 2015-04-14
Token naming refactor

mattheworiordan:
This PR tackles all the refactoring discussed in https://github.com/ably/ably-js/issues/33#issuecomment-92515029 with the exception that any field name or type changes on the server-side have not been changed but instead have been marked with `TODO`.

@kouno and @paddybyers I believe this covers pretty much everything so can be used as the basis for discussions with client library developers in regards to supporting this updated API.

We should discuss the few remaining points starting at https://github.com/ably/ably-js/issues/33#issuecomment-92546585 so that we can hopefully finalise and lock down the API now.

I've just realised that this list of commits includes the `until_attach` PR, so the changes for this PR only are in https://github.com/ably/ably-ruby/compare/4425f4e...token-naming-refactor


mattheworiordan:
BTW. https://github.com/ably/ably-ruby/compare/4425f4e...token-naming-refactor contains the commits as [until_attach PR](https://github.com/ably/ably-ruby/pull/22) is not merged yet.  Can you please feedback on that too so that I can merge if you are happy?

PR #30, opened 2015-04-15
Use :auth_callback instead of a block for consistency

mattheworiordan:
In the initial implementation of the Ruby client library a block was used for the auth callback.

This is inconsistent with the other client libraries, so it is now a `:auth_callback` option.

@paddybyers and @kouno FYI.  I'd like to merge ASAP.


mattheworiordan:
Merged in https://github.com/ably/ably-ruby/commit/b851323fd115cf183c411293af90bf1c6f8daef6

PR #48, opened 2015-05-12
Reject invalid payload type

mattheworiordan:
@SimonWoolf & @paddybyers following our discussion today, commit c3309fd ensures that non supported data payload types raise an exception immediately.

I had to add a new error code too, see https://github.com/ably/ably-common/commit/5ea9a3334309f9e108f1bed2ffff09c70043a043

Can you tell me if you are happy with this?  If so, I will amend our [feature documentation](http://docs.ably.io/client-lib-development-guide/features/) and communicate this out to everyone.

FYI @kouno

### DIFF
--- a/lib/ably/exceptions.rb
+++ b/lib/ably/exceptions.rb
@@ -74,5 +74,8 @@ module Ably
 
     # The message could not be delivered to the server
     class MessageDeliveryError < BaseAblyException; end
+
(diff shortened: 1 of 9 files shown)


SimonWoolf:
Corresponding PR for ably-js: https://github.com/ably/ably-js/pull/53

PR #51, opened 2015-05-27
Ensure files are required in a consistent order

SimonWoolf:
The order files are returned by Dir.glob is [unspecified and OS-dependent](http://ruby-doc.org/core-2.2.2/Dir.html#method-c-glob), which was causing odd behaviour on Linux. Add a sort to ensure files are required in a consistent order across all OSes


mattheworiordan:
Nice, thanks

PR #58, opened 2015-09-02
Allow clientId to be provided on init if using externally created token

SimonWoolf:
Thought I'd add a test to check ably-ruby didn't have the problem of https://github.com/ably/ably-js/issues/108 , needed a tweak to make it pass


mattheworiordan:
Looks good :+1:

PR #78, opened 2016-01-26
Fix ruby verbose warnings

mattheworiordan:
@lmars going to merge this ASAP unless you have any objections.

This will remove all the warnings from Ruby when in verbose mode


lmars:
LGTM

PR #81, opened 2016-02-28
Encryption spec update

mattheworiordan:
I've gone ahead and rolled out the changes to the spec in Ruby as well which fixes https://github.com/ably/ably-ruby/issues/80

I discovered a few bugs along the way so fixed everything that needed doing before implementing the Crypto changes.

cc @SimonWoolf

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,7 +157,7 @@ channel.presence.history do |presence_page|
 end
 ```
 
-### Symmetrical end-to-end encrypted payloads on a channel
+### Symmetric end-to-end encrypted payloads on a channel
 
 When a 128 bit or 256 bit key is provided to the library, all payloads are encrypted and decrypted automatically using that key on the channel. The secret key is never transmitted to Ably and thus it is the developer's responsibility to distribute a secret key to both publishers and subscribers.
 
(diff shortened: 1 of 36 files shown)


README.md
```diff
@@ -157,6 +157,19 @@ channel.presence.history do |presence_page|
...
+
+```ruby
+secret_key = Ably::Util::Crypto.generate_random_key
+channel = client.channels.get('test', cipher: { key: secret_key })
+channel.subscribe do |message|
+  message.data #=> "sensitive data"
+end
+channel.publish nil, "sensitive data" # data will be encrypted before publish
+```
+
```
SimonWoolf: Suggestion: could make it clear that the name isn't encrypted? eg
```ruby
secret_key = Ably::Util::Crypto.generate_random_key
channel = client.channels.get('test', cipher: { key: secret_key })
channel.subscribe do |message|
  message.data #=> "sensitive data (encrypted before being published)"
end
channel.publish "name (not encrypted)", "sensitive data (encrypted before being published)"
```

SimonWoolf:
LGTM :)

mattheworiordan:
> Is a combined option something that we want to support in general? (I haven't included it in ably-js, and it isn't in the spec)

I don't think it needs to be part of the spec, I opted to include it in the Ruby lib

README.md
mattheworiordan: Sure, will do

README.md
```diff
@@ -157,6 +157,19 @@ channel.presence.history do |presence_page|
 end
 ```
 
+### Symmetrical end-to-end encrypted payloads on a channel
```
paddybyers: I think it should be symmetric encryption, not symmetrical.

mattheworiordan: Yup thanks

PR #84, opened 2016-03-16
Proxy support

mattheworiordan:
@paddybyers see my initial implementation for Proxy support, REST works perfectly.  Unfortunately the [WebSocket](https://www.ably.io/concepts/websockets) library being used does not seem to support proxies correctly, so I've raised an issue at https://github.com/faye/websocket-driver-ruby/issues/40.

### DIFF
--- a/README.md
+++ b/README.md
@@ -227,10 +227,10 @@ secret_key = Ably::Util::Crypto.generate_random_key
 channel = client.channels.get('test', cipher: { key: secret_key })
 channel.publish nil, "sensitive data" # data will be encrypted before publish
 messages_page = channel.history
-messages_page.first.data #=> "sensitive data"
+messages_page.items.first.data #=> "sensitive data"
 ```
 
-### Generate Token
+### Generate a Token
 
 Tokens are issued by Ably and are readily usable by any client to connect to Ably:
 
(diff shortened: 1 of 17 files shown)


mattheworiordan:
@paddybyers I have now completed proxy support for both the REST and Realtime Ruby libraries.  However, I have been unable to get Digest authentication to work, and looking into this, I think it would unfortunately require significant changes to the underlying websocket library which I do not have the appetite to do at this stage.  Personally I think we could go without Digest for now, and if easy to add in other libraries, then we can add support as necessary.

Please note the API I have chosen is:

```ruby
ably = Ably.Realtime.new(key: key, proxy: { host: "x", port: 99, username: "y", password: "z")
```

lukaszsliwa:
This pull-request changes `websocket-driver` to `faye-websocket-driver`. I am not sure if it was an intention to do it as the `main` branch does not have such change.
`http` proxy protocol is hardcoded in the proxy url, so `https`, `ws`, `wss` protocols will **not** work with this pull request. It would be better to add `protocol` param too, or to set string url to the proxy parameter i.e.:

```ruby
Ably::Realtime::Client.new(key: '...', proxy: 'http://username:password@my.proxy.dev:1234/')
```

This is how`Faraday` implements it and no need to prepare `proxy_url`

I did not check states but there are some changes in connection transporter and log_level (replaced `fatal` with `error`). I couldn't make it work after merging `main` into this pull-request. Specs fails.

PR #94, opened 2016-08-08
Ensure interoperability with other libraries with JSON protocol

mattheworiordan:
@paddybyers @tcard FYI, this is my implementation of interoperability with Ruby.

Note that:

* These tests explicitly don't use MsgPack as the encodings are different. I am not sure if we want separate tests for MsgPack, I would assume we do, but much harder to do as a raw test I would imagine.
* This depends on some updates I made in https://github.com/ably/ably-common/pull/18, specifically the `expectedHexValue` attribute.

See https://github.com/ably/ably-ios/pull/459 and https://github.com/ably/ably-common/pull/16

### DIFF
--- a/spec/acceptance/realtime/message_spec.rb
+++ b/spec/acceptance/realtime/message_spec.rb
@@ -739,49 +739,115 @@ describe 'Ably::Realtime::Channel Message', :event_machine do
 
   context 'message encoding interoperability' do
     let(:client_options)  { { key: api_key, environment: environment, protocol: :json } }
-    let(:realtime_client) do
-      auto_close Ably::Realtime::Client.new(client_options)
-    end
-    let(:rest_client) do
-      Ably::Rest::Client.new(client_options)
-    end
     let(:channel_name) { "subscribe_send_text-#{random_str}" }
-    let(:realtime_channel) { realtime_client.channels.get(channel_name) }
 
     fixtures_path = File.expand_path('../../../../lib/submodules/ably-common/test-resources/messages-encoding.json', __FILE__)
 
-    JSON.parse(File.read(fixtures_path))['messages'].each do |encoding_spec|
-      context "when decoding #{encoding_spec['expectedType']}" do
-        it 'ensures that client libraries have compatible encoding and decoding using common fixtures' do
-          realtime_channel.attach do
-            realtime_channel.subscribe do |message|
-              if encoding_spec['expectedHexValue']
-                expect(message.data.unpack('H*').first).to eql(encoding_spec['expectedHexValue'])
+    context 'over a JSON transport' do
+      let(:realtime_client) do
+        auto_close Ably::Realtime::Client.new(client_options)
+      end
+      let(:rest_client) do
+        Ably::Rest::Client.new(client_options)
+      end
+      let(:realtime_channel) { realtime_client.channels.get(channel_name) }
+
+      JSON.parse(File.read(fixtures_path))['messages'].each do |encoding_spec|
+        context "when decoding #{encoding_spec['expectedType']}" do
+          it 'ensures that client libraries have compatible encoding and decoding using common fixtures' do
+            realtime_channel.attach do
+              realtime_channel.subscribe do |message|
+                if encoding_spec['expectedHexValue']
+                  expect(message.data.unpack('H*').first).to eql(encoding_spec['expectedHexValue'])
+                else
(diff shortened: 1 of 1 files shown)


tcard:
LGTM, but the submodule pointer is pointer to a fixup commit that will probably disappear, so it should be updated to a stable one before merging.

mattheworiordan:
> LGTM, but the submodule pointer is pointer to a fixup commit that will probably disappear, so it should be updated to a stable one before merging.

Yup, will wait until it'a all merged to master

mattheworiordan:
@tcard @paddybyers please review 3d660cb which ensures that content is compared with key/value pairs and array objects in no specific order, and also now introduces MsgPack interoperability tests.  Output from test suite is as follows:

```
Ably::Realtime::Channel Message
  message encoding interoperability
    over a MsgPack transport
      when publishing a string using JSON protocol
        receives the message over MsgPack and the data matches
      when retrieving a string using JSON protocol
        is compatible with a publishes using MsgPack
      when retrieving a jsonArray using JSON protocol
        is compatible with a publishes using MsgPack
      when publishing a binary using JSON protocol
        receives the message over MsgPack and the data matches
      when publishing a string using JSON protocol
        receives the message over MsgPack and the data matches
      when publishing a jsonArray using JSON protocol
        receives the message over MsgPack and the data matches
      when retrieving a jsonObject using JSON protocol
        is compatible with a publishes using MsgPack
      when retrieving a binary using JSON protocol
        is compatible with a publishes using MsgPack
      when retrieving a string using JSON protocol
        is compatible with a publishes using MsgPack
      when publishing a jsonObject using JSON protocol
        receives the message over MsgPack and the data matches
    over a JSON transport
      when decoding string
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when encoding string
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when encoding jsonArray
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when decoding binary
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when decoding string
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when decoding jsonArray
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when encoding jsonObject
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when encoding binary
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when encoding string
        ensures that client libraries have compatible encoding and decoding using common fixtures
      when decoding jsonObject
        ensures that client libraries have compatible encoding and decoding using common fixtures
```

If you are happy with this, I will write up a new spec item for 0.9

paddybyers:
I don't understand the code really but you're not saying that order is not significant in array members?

mattheworiordan:
@paddybyers on Array ordering, [I did ignore ordering](https://github.com/ably/ably-ruby/pull/94/files#diff-5a92463d6ab8ad3fc8cf8aff5cfde6f8R746), not sure why I did that.  Fixed in e835f8d

tcard:
Small comment, LGTM otherwise. I'll replicate this on https://github.com/ably/ably-ios/pull/459 and https://github.com/ably/ably-java/pull/136.

spec/acceptance/realtime/message_spec.rb
```diff
@@ -736,4 +736,128 @@ def self.add_tests_for_data(data)
...
+              encoding_spec['expectedHexValue'].scan(/../).map { |x| x.hex }.pack('c*')
+            else
+              encoding_spec['expectedValue']
+            end
+
+            realtime_channel.publish("event", data) do
+              response = rest_client.get("/channels/#{channel_name}/messages")
+              message = response.body[0]
+              expect(message['encoding']).to eql(encoding_spec['encoding'])
+              expect_data_to_match encoding_spec['data'], message['data']
```
tcard: This is where order matters, as those values are JSON-encoded maps and can be like `{"a":123,"b":456}` and `{"b":456,"a":123}` and would fail to match given that they are `String`s. (Ruby seems to do what we expect, ie. to keep the order from the fixture, so no big deal I think.)

mattheworiordan: Good point, I didn't go back and look at this previous test.  For completeness I will check the encoding and make sure that ordering is ignored.

mattheworiordan: See 749b246

PR #99, opened 2016-10-02
0.9 master

mattheworiordan:
WIP branch we will use to track progress on 0.9 for realtime spec adherence.


mattheworiordan:
@SimonWoolf @paddybyers I believe `ably-ruby` is now feature complete for the 0.9 spec with two caveats:

* I just noticed `RTN13e` (ping with ID) is not yet implemented, I'll do that later, it's a tiny change.
* I want to check each method where I am raising exceptions instead of using the Deferrable error callback for methods and make sure that exceptions are only raised when the arguments are incorrect and invalid in all cases, whereas if the method fails because of the current client library state, the error callback is used instead so that developers have a predictable way to catch failures in all states.  I touched on this in https://github.com/ably/wiki/issues/175, and think it's important for all libraries.

Just waiting for Travis to finish building too in case there are any other broken tests from the changes...

mattheworiordan:
Also, we need to agree on https://github.com/ably/docs/pull/238 as I have implemented those proposed changes

mattheworiordan:
@SimonWoolf @paddybyers ably-ruby is now feature complete and I've knocked off all outstanding issues too.

mattheworiordan:
Superseded by #111

PR #118, opened 2017-05-11
Encode presence URI correctly

smeyfroi:
Avoids this error when doing ```channel.presence.get``` or ```channel.presence.history```:
```
Addressable::URI::InvalidURIError: Invalid scheme format: 00000000
```

PR #115, opened 2017-04-11
Push notifications

mattheworiordan:
This PR adds push notification support to the Ably Ruby library.

### Installation instructions

Add this line to your application's Gemfile:

```
gem 'ably', git: 'https://github.com/ably/ably-ruby', branch: 'push-notifications'
```

And then install this Bundler dependency:

```
$ bundle
```

### Outstanding tasks for push notification support:

* [x] Realtime wrapper APIs around REST APIs for admin functionality
* [x] Channel subscriptions
* [x] Get skipped tests running once realtime issues fixed

PR #119, opened 2017-05-11
Channel name encoding error for REST requests

mattheworiordan:
This PR is just [@smeyfroi's PR](https://github.com/ably/ably-ruby/pull/118) plus one additional test to prevent a regression.

### DIFF
--- a/lib/ably/rest/channel.rb
+++ b/lib/ably/rest/channel.rb
@@ -130,7 +130,7 @@ module Ably
 
       private
       def base_path
-        "/channels/#{Addressable::URI.encode(name)}"
+        "/channels/#{URI.encode_www_form_component(name)}"
       end
 
       def decode_message(message)
--- a/lib/ably/rest/presence.rb
+++ b/lib/ably/rest/presence.rb
@@ -84,7 +84,7 @@ module Ably
 
       private
       def base_path
-        "/channels/#{Addressable::URI.encode(channel.name)}/presence"
+        "/channels/#{URI.encode_www_form_component(channel.name)}/presence"
       end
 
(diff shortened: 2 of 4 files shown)


smeyfroi:
pr looks good but travis not happy... these look like the unrelated fails i was seeing on the old pr... need to leave these to you @mattheworiordan i reckon, sorry!

mattheworiordan:
> pr looks good but travis not happy.

Good news is that those errors are unrelated and not an issue.  One is a test issue introduced by a hotfix https://github.com/ably/ably-ruby/commit/ee038c398fb91bd2fed55f1cafcc498bad178179.  The others are errors simply due to the fixtures not working as expected for push notifications and this test suite not supporting push (see https://github.com/ably/ably-ruby/pull/115).  So this branch is definitely save for you to use i.e. in your Gemfile `gem `ably`, git: 'https://github.com/ably/ably-ruby.git', branch: 'presence_error'`.

I will get this fixed along with the heartbeat issue soon and do a new release.

smeyfroi:
ah great. i have the 'ably-rest' gem in my bundle here, which doesn't do much beyond bringing this repo in as a submodule. can i just use 'ably-ruby' in my bundle if i'm only using the rest api?

mattheworiordan:
> can i just use 'ably-ruby' in my bundle if i'm only using the rest api?

Yup, you can, you just get a few extra dependencies but it won't make any difference in reality.  I'll try and get a release out next week.

smeyfroi:
awesome, thanks matthew :-)

mattheworiordan:
This is now merged and available in stable release v1.0.4

PR #122, opened 2017-08-02
Make it easier to run ruby tests against a local realtime

SimonWoolf:


SimonWoolf:
Closing this, way too many other places in the test suite that create a new client from scratch and don't use TestApp ports, it's easier just to do `sudo socat tcp-listen:80,reuseaddr,fork tcp:localhost:8080` and `sudo socat tcp-listen:443,reuseaddr,fork tcp:localhost:8081`...

PR #133, opened 2018-04-17
Add request_id option to client

funkyboy:


lib/ably/rest/client.rb
```diff
@@ -146,6 +150,7 @@ def initialize(options)
         @custom_host      = options.delete(:rest_host)
         @custom_port      = options.delete(:port)
         @custom_tls_port  = options.delete(:tls_port)
+        @add_request_ids  = options.delete(:add_request_ids) == false ? false : true
```
SimonWoolf: As an optional extra debug option, this should default to false (unless and until it's added to the spec). So just `@add_request_ids  = options.delete(:add_request_ids)` is fine.

(Also, `cond ? false : true` is a pointless pattern, it's equivalent to `!cond`)

lib/ably/rest/client.rb
```diff
@@ -434,6 +439,16 @@ def send_request(method, path, params, options)
...
+        if @add_request_ids
+          if method == :get
+            path_with_params = Addressable::URI.new
+          else
+            path_with_params = Addressable::URI.parse(path)
+          end
+          path_with_params.query_values = {request_id: SecureRandom.hex(20)}
+          query = path_with_params.query
+          path = "#{path}?#{query}"
+        end
```
SimonWoolf: Why are you manually munging the path? This method takes a params hash. You can add it to the params hash.

spec/unit/rest/client_spec.rb
```diff
@@ -50,4 +50,11 @@
       end
     end
   end
+
+  context 'request_id generation' do
+    let(:client_options) { {key: 'appid.keyuid:keysecret', add_request_ids: true} }
+    it 'includes request_id in URL' do
+      expect(subject.add_request_ids).to eql(true)
+    end
+  end
```
SimonWoolf: If you're going to add a test, may as well be one that checks the result (i.e. uses webmock to check that a `request_id` actually gets added to the querystring for rest requests, and checks timeout exceptions to make sure that the request ID gets printed with those), rather than just checks that an instance variable got set correctly, no?

funkyboy:
@SimonWoolf re-ready for review.

lib/ably/exceptions.rb
```diff
@@ -13,9 +13,9 @@ module Exceptions
     # @!attribute [r] code
     #   @return [String] Ably specific error code
     class BaseAblyException < StandardError
-      attr_reader :status, :code
+      attr_reader :status, :code, :request_id
 
-      def initialize(message, status = nil, code = nil, base_exception = nil, options = {})
+      def initialize(message, status = nil, code = nil, base_exception = nil, options = {}, request_id = nil)
```
mattheworiordan: I would have preferred for `request_id` to be an option given it's a flexible hash for this reason.

lib/ably/exceptions.rb
```diff
@@ -71,8 +72,8 @@ class EncoderError < BaseAblyException; end
 
     # Connection error from Realtime or REST service
     class ConnectionError < BaseAblyException
-      def initialize(message, status = nil, code = nil, base_exception = nil, options = {})
-        super message, status, code, base_exception, options
+      def initialize(message, status = nil, code = nil, base_exception = nil, options = {}, request_id = nil)
+        super message, status, code, base_exception, options, request_id
```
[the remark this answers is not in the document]
mattheworiordan: If we leave `request_id` as an option of options, then no change here is needed.

lib/ably/rest/client.rb
```diff
@@ -434,6 +439,10 @@ def send_request(method, path, params, options)
         max_retry_duration = http_defaults.fetch(:max_retry_duration)
         requested_at       = Time.now
         retry_count        = 0
+        if @add_request_ids
```
mattheworiordan: It's bad practice to read an instance variable like this when it's available as a reader i.e. `if @add_request_idds` would always be false due to a type, whereas `if add_request_idds` would fail.

lib/ably/rest/client.rb
```diff
@@ -456,12 +465,12 @@ def send_request(method, path, params, options)
             logger.warn { "Ably::Rest::Client - Retry #{retry_count} for #{method} #{path} #{params} as initial attempt failed: #{error}" }
             retry
           end
-
+          request_id = params[:request_id] if @add_request_ids
           case error
             when Faraday::TimeoutError
-              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error)
+              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, {}, request_id)
```
mattheworiordan: Based on my proposed change, it would be `raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, { request_id: request_id })`.
I think `raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, request_id: request_id)` may work too, although I recall 1.9.2 having some constraints with pseudo named args.

spec/unit/rest/client_spec.rb
```diff
@@ -50,4 +50,11 @@
       end
     end
   end
+
+  context 'request_id generation' do
+    let(:client_options) { {key: 'appid.keyuid:keysecret', add_request_ids: true} }
```
mattheworiordan: Unusual spacing, in all other places `{` has a space after, and `}` a space before

lib/ably/exceptions.rb
funkyboy: Didn’t know the supposed usage of that hash, that’s why I didn’t touch it. Good to know that I can use it.

funkyboy:
@SimonWoolf @mattheworiordan 

- added `request_id` for nonce errors
- now passing `request_id` in the `options` param
- added more tests for both nonce and timeout errors with `add_request_ids` true and false

Ready for another look.

spec/acceptance/rest/client_spec.rb
```diff
@@ -971,6 +971,20 @@ def encode64(text)
...
+      context 'Nonce error' do
+        let(:token_params) { { nonce: "samenonce_#{protocol}", timestamp:  Time.now.to_i} }
+        it 'includes request_id in UnauthorizedRequest error due to replayed nonce' do
+          client1 = Ably::Rest::Client.new(default_options.merge(key: api_key))
+          client2 = Ably::Rest::Client.new(default_options.merge(key: api_key, add_request_ids: true))
+          expect { client1.auth.request_token(token_params) }.not_to raise_error
+          begin
+            client2.auth.request_token(token_params)
+          rescue Ably::Exceptions::UnauthorizedRequest => err
+            expect(err.request_id).to_not eql(nil)
```
mattheworiordan: Do we have any test to ensure that when `add_request_ids` is false, that the `request_id` is `nil`?

lib/ably/rest/client.rb
mattheworiordan: @SimonRichardson because this is Ruby and not Elixir.  So the `params` object would be changed by just calling this method, which is bad.

lib/ably/rest/client.rb
```diff
@@ -470,9 +470,9 @@ def send_request(method, path, params, options)
           end
           case error
             when Faraday::TimeoutError
-              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, {}, random_request_id)
+              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, {request_id: random_request_id})
```
mattheworiordan: Spacing issues again, here and many times below

mattheworiordan: I am not a big fan of this. `random_request_id` is defined in line 445, yet is within a conditional.  I realise Ruby allows this sort of thing, but it's messy and error prone.  Why are you defining `random_request_id` as a varaiable when you have access to the `params` variable in this scope which is available anyway?

mattheworiordan: In fact, I am not sure I have seen a test that confirms that a) a request is sent to Ably with a `request_id`, b) it fails (intentionally), c) the test confirms that the exception raised has that SAME `request_id`, d) the Logger is showing the `request_id` in the output.  This is the crux of what we're doing here.  Given this code path, I fear our tests so far are just testing that a `request_id` is present, but not ensuring it is consistent throughout.

lib/ably/rest/middleware/exceptions.rb
```diff
@@ -27,7 +27,7 @@ def on_complete(env)
             end
 
             message = 'Unknown server error' if message.to_s.strip == ''
-            request_id = env.request.context[:request_id]
+            request_id = env.request.context[:request_id] if env.request.context
```
mattheworiordan: Spacing issues below.

lib/ably/rest/client.rb
```diff
@@ -456,12 +465,12 @@ def send_request(method, path, params, options)
...
+          request_id = params[:request_id] if @add_request_ids
           case error
             when Faraday::TimeoutError
-              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error)
+              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, {}, request_id)
             when Faraday::ClientError
-              raise Ably::Exceptions::ConnectionError.new(error.message, nil, 80000, error)
+              raise Ably::Exceptions::ConnectionError.new(error.message, nil, 80000, error, {}, request_id)
             else
               raise error
```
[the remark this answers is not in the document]
mattheworiordan: I am also not keen on a var called `random_request_id`, that sounds more like a method.  It's no longer random once it's assigned, and thus it's consistent for the request.

lib/ably/rest/client.rb
SimonWoolf: > So the params object would be changed by just calling this method, which is bad.

@mattheworiordan fair enough, but imo the solution to that is to clone the hash (as Cesare has done in the latest version of this pr), not to start munging a param into the path (meaning faraday has to parse it out again to combine with the params -- assuming it even does that rather than give you a path with two `?`s). (for one thing the original approach assumes the path doesn't already have a querysting, so fails the 'what if two things did this' test)

spec/acceptance/rest/client_spec.rb
```diff
@@ -963,16 +963,17 @@ def encode64(text)
...
+              client.stats
+            rescue Ably::Exceptions::ConnectionTimeout => err
+              expect(err.request_id).to_not eql(nil)
+            end
         end
       end
-    end
-        context 'without request_id' do
+
+      context 'without request_id' do
```
mattheworiordan: This doesn't fix the issue?  @funkyboy what is happening here?  Can you explain what the problem is as this is the second attempt to fix indentation issues and it's failed.  What IDE are you using?

capto_capture 2018-04-18_06-24-58_pm

lib/ably/rest/client.rb
funkyboy: This makes sense, but I am struggling a bit to inspect the url of a request, say `client.stats`. Any suggestion?

spec/acceptance/rest/client_spec.rb
funkyboy: I am using VS code but clearly there’s something wrong going on. Will open an issue to add rubocop, so we don’t waste time on this.

funkyboy:
- [ ] Fix indentation (for real)
- [ ] Add test to verify that the `request_id` is the same at request and response time, and that it's included in the error and the logger
- [ ] Add test to verify that, when a request fails, it is retried to a different endpoint with the same `request_id`. See https://github.com/ably/docs/issues/359

lib/ably/rest/client.rb
```diff
@@ -434,11 +439,22 @@ def send_request(method, path, params, options)
         max_retry_duration = http_defaults.fetch(:max_retry_duration)
         requested_at       = Time.now
         retry_count        = 0
+        request_id         = nil
+        if add_request_ids
+          params = {} if params.nil?
```
mattheworiordan: This is a tiny comment, but it seems unnecesary to dup an empty hash, so perhaps instead just:

```ruby
params = if params.nil?
  {}
else
  params.dup
end
```

lib/ably/rest/client.rb
```diff
@@ -456,12 +472,11 @@ def send_request(method, path, params, options)
...
             retry
           end
-
           case error
             when Faraday::TimeoutError
-              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error)
+              raise Ably::Exceptions::ConnectionTimeout.new(error.message, nil, 80014, error, { request_id: request_id })
             when Faraday::ClientError
-              raise Ably::Exceptions::ConnectionError.new(error.message, nil, 80000, error)
+              raise Ably::Exceptions::ConnectionError.new(error.message, nil, 80000, error, { request_id: request_id })
```
mattheworiordan: May be nice for future us to add a note about the `raise error` not explicitly passing the `request_id`, but in the Middleware it's available through the requext context. WDYT?

PR #135, opened 2018-04-18
Fix failing auth test

funkyboy:
There's a new `40104` code for invalid timestamp. 
This fixes the related ruby test.


mattheworiordan:
👍

PR #141, opened 2018-04-23
Fix presence history test

funkyboy:
Couldn't come up with a better idea at the moment :(


funkyboy:
@mattheworiordan you mean something like https://github.com/ably/ably-ruby/pull/141/commits/20239cf29ea3439c0e4fecc4b72848347208b0f5 ?

PR #146, opened 2018-04-27
Only resume if connection is fresh (RTN15g*)

mattheworiordan:
Previously we would resume a connection even if the last time we had contacted Ably was greater than the max connection state TTL. As a result, we would try and resume a connection that is guaranteed not to be there.

This implements spec at https://github.com/ably/docs/pull/331

### DIFF
--- a/lib/ably/realtime/connection.rb
+++ b/lib/ably/realtime/connection.rb
@@ -615,6 +615,8 @@ module Ably
       def connection_state_available?
         return true if connected?
 
+        return false if time_since_connection_confirmed_alive? > connection_state_ttl
+
         connected_last = state_history.reverse.find { |connected| connected.fetch(:state) == :connected }
         if connected_last.nil? || (connected_last.fetch(:transitioned_at) < Time.now - connection_state_ttl)
           false
--- a/spec/acceptance/realtime/connection_failures_spec.rb
+++ b/spec/acceptance/realtime/connection_failures_spec.rb
@@ -514,6 +514,173 @@ describe Ably::Realtime::Connection, 'failures', :event_machine do
           end
         end
 
+        context 'connection state freshness is monitored' do
+          it 'resumes connections when disconnected within the connection_state_ttl period (#RTN15g)' do
+            connection.once(:connected) do
+              connection_id = connection.id
+              reconnected_with_resume = false
+
+              # Make sure the next connect has the resume param
+              allow(EventMachine).to receive(:connect).and_wrap_original do |original, *args, &block|
+                url = args[4]
+                uri = URI.parse(url)
+                expect(CGI::parse(uri.query)['resume'][0]).to_not be_empty
+                reconnected_with_resume = true
+                original.call(*args, &block)
+              end
+
+              connection.once(:disconnected) do
+                disconnected_at = Time.now
+
+                connection.once(:connecting) do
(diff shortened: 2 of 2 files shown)


lib/ably/realtime/connection.rb
```diff
@@ -615,6 +615,8 @@ def connection_resumable?
       def connection_state_available?
         return true if connected?
 
+        return false if time_since_connection_confirmed_alive? > connection_state_ttl
```
funkyboy: Shouldn't this check include also the idle interval? Much like JS does https://github.com/ably/ably-js/blob/c55b4518d6f096e4bde90afcd04c5ba35d466f3c/common/lib/transport/connectionmanager.js#L706

lib/ably/realtime/connection.rb
```diff
@@ -615,6 +615,8 @@ def connection_resumable?
...
+
         connected_last = state_history.reverse.find { |connected| connected.fetch(:state) == :connected }
         if connected_last.nil? || (connected_last.fetch(:transitioned_at) < Time.now - connection_state_ttl)
           false
```
SimonWoolf: I know not part of this PR, but that doesn't look right -- you don't resume if the most recent time you transitioned to­ (not from) `connected` was more than connectionStateTtl ago? Am I reading that correctly?

lib/ably/realtime/connection.rb
mattheworiordan: Oops, yes, I forgot the idle interval. Thanks, good spot, I was being a bit too strict there.

lib/ably/realtime/connection.rb
mattheworiordan: Yes, you are reading it incorrectly, it's time since connection comfirmed alive i.e. freshness.  This existing method is used to detect a connection that is no longer active.

SimonWoolf: > Yes, you are reading it incorrectly, it's time since connection comfirmed alive i.e. freshness. 

No, that comment is on lines 619-620 (hence 'I know not part of this PR'). This bit: `if connected_last.nil? || (connected_last.fetch(:transitioned_at) < Time.now - connection_state_ttl)`

> This existing method is used to detect a connection that is no longer active.

Yes, I can see that from the method name, I'm just not sure the logic is correct.

SimonWoolf: To be explicit: if I'm reading this correctly, line 621 says "don't resume if the time of the last transition to `connected` is over 2 minutes ago". So if you're connected for an hour, then your connection drops for 5 seconds, it won't resume because the last transition to `connected` was an hour ago.

mattheworiordan: Mmm, I didn't look at that code, but that does look suspect. I will investigate.

mattheworiordan: @SimonWoolf btw. I pushed b70811e earlier, will see what that breaks :)

mattheworiordan:
@funkyboy @SimonWoolf build is 🍏 https://travis-ci.org/ably/ably-ruby/builds/372091535
Happy for me to merge this + encoded strings fix https://github.com/ably/ably-ruby/pull/147.  @SimonWoolf we could perhaps do a release once your PR is ready?

PR #140, opened 2018-04-19
Do not encode strings in-place

aschuster3:
Pretty straight forward fix.  As Ruby moves towards making all string literals frozen by default, more people might hit problems with this as time goes on.

Fixes https://github.com/ably/ably-ruby/issues/132

PR #147, opened 2018-04-27
Do not encode strings in-place

mattheworiordan:
This builds on the contribution from @aschuster3 at https://github.com/ably/ably-ruby/pull/140.

Hopefully this additional commit will ensure that returned values are stored as instance variables or in the scope they are intended without side-effects now.  Will wait for CI to see 🍾


mattheworiordan:
@aschuster3 @funkyboy the build has passed, see https://travis-ci.org/ably/ably-ruby/builds/371997475.

I propose we merge this now?

Issue #126, opened 2018-01-24
Check fallback functionality is being used after timeouts


SimonWoolf:
Customer reporting getting occasional `Faraday::ConnectionFailed` and `Net::ReadTimeout` errors with REST publishes. On timeout errors the lib should be retrying the request (to a different datacenter) per RSC15d, and it should have time to make at least a second attempt no matter what class of timeout is encountered on the first attempt per TO3l4 and TO3l6. Need to verify that this happens for both those classes of timeout errors, and fix it if not.

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4078) by [Unito](https://www.unito.io)

mattheworiordan:
> Need to verify that this happens for both those classes of timeout errors, and fix it if not.

I am not sure there is any evidence this is not happening, and as far as I can tell, so long as the ClientOptions are configured correctly, this is guaranteed.  There are tests for this too https://github.com/ably/ably-ruby/blob/master/spec/acceptance/rest/client_spec.rb#L571-L614

SimonWoolf:
>  There are tests for this too 

There are, but at a closer look I'm not quite convinced by them. e.g. they pass a `max_retry_duration` param, but that's just ignored as the lib expects it to be `http_max_retry_duration`. Which is why the second test is passing even though it expects the opposite of what should happen (ie it [asserts that a fallback](https://github.com/ably/ably-ruby/blob/master/spec/acceptance/rest/client_spec.rb#L611) was tried even though it's passing options that should mean no fallbacks are tried). Even after that's fixed, fallbacks seem to be completely broken for GET, probably due to https://bugs.ruby-lang.org/issues/10674

That said, except for the GET issue those are all test bugs, and at least for POST the lib seems to be working as expected afaict -- and the requests the customer was complaining about are POSTs rather than GETs. So happy for you to close this.

Pushed some fixed (and so failing for GETs) tests up to https://github.com/ably/ably-ruby/compare/fallback-on-timeout-spec?expand=1

PR #148, opened 2018-04-27
RSC15d test fixes; add (failing) tests for GET as well as POST

SimonWoolf:
Linked these from https://github.com/ably/ably-ruby/issues/126#issuecomment-382725426, but said I'd make a pr. Main issue with the previous tests is that they pass a `max_retry_duration` param, but that's just ignored as the lib expects it to be `http_max_retry_duration`, which is why the old second test was passing even though it expects the opposite of what should happen (it asserts that a fallback was tried even though it's passing options that should mean no fallbacks are tried).

The GET test I added is failing, probably due to https://bugs.ruby-lang.org/issues/10674 . No attempt at a fix here, just highlighting the issue.


SimonWoolf:
> Do you think with dynamic languages we should perhaps warn people when they're passing in options that are not supported in Hashes/Dictionary/Object-like arguments?

Sure, why not, seems reasonable

mattheworiordan:
2 tests are consistently failing in this PR...

SimonWoolf:
> 2 tests are consistently failing in this PR...

Yes. See original comment.

>> The GET test I added is failing, probably due to https://bugs.ruby-lang.org/issues/10674 . No attempt at a fix here, just highlighting the issue.

mattheworiordan:
@SimonWoolf @paddybyers @funkyboy you are going to love this one.  I thought whilst I am deep in Ruby code, I may as well just fix @SimonWoolf's PR as it had some failing tests.

After a little investigation, I noticed that the local web server was reporting two requests being received, when I was 100% sure that the client library was making only one request.  And then I discovered that Ruby retries what it believes are idempotent requests twice without providing any feedback...  See https://github.com/ably/ably-ruby/pull/148/commits/cb32020dceaa096cc31027f57e6e908afde0fb2e.

To be honest I am not 100% convinced we should introduce a whole new HTTP library to this Gem just to work around the problem, but it does seem like we have little choice unfortunately before Ruby 2.5 to fix this problem.

WDYT?

mattheworiordan:
@SimonWoolf oops, I should read your comments 👍

PR #150, opened 2018-04-28
Add .editorconfig for basic IDE configuration settings

mattheworiordan:
See http://editorconfig.org/ and https://github.com/ably/docs/issues/427

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+# http://EditorConfig.org
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent-size = 2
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,14 @@
...
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent-size = 2
+charset = utf-8
+
+[*.md]
+trim_trailing_whitespace = false
```
funkyboy: Just a curiosity. Why is it false for markdown files?

mattheworiordan: Good question.  I don't know, I saw this in a few other examples I looked at.  I can only imagine that for things like code blocks, you may for some reason want to force additional trailing white space.  But given I can't think of a good reason for this, I should really remove.

PR #165, opened 2018-07-22
Idempotent publishing

mattheworiordan:
Unless I am mistaken, we don't have a spec for idempotent publishing, so I have created this PR to help us define what functionality we may to include in the spec for this. See issue at https://github.com/ably/docs/issues/469 relating to idempotent publishing in the spec.

Note:

* I additionally added a new ClientOptions as a beta feature that enables idempotent publishing to all publishes.  It's disabled by default. I propose this is included in 1.1.
* When we release 1.2, we may want to consider enabling this feature by default.

### DIFF
--- a/lib/ably/rest/channel.rb
+++ b/lib/ably/rest/channel.rb
@@ -22,6 +22,8 @@ module Ably
       # @api private
       attr_reader :push
 
+      IDEMPOTENT_LIBRARY_GENERATED_ID_LENGTH = 9 # See spec RSL1k1
+
       # Initialize a new Channel object
       #
       # @param client [Ably::Rest::Client]
@@ -93,7 +95,7 @@ module Ably
             # We cannot mutate for idempotent publishing if one or more messages already has an ID
             if payload.all? { |msg| !msg['id'] }
               # Mutate the JSON to support idempotent publishing where a Message.id does not exist
-              idempotent_publish_id = SecureRandom.base64(16)
+              idempotent_publish_id = SecureRandom.base64(IDEMPOTENT_LIBRARY_GENERATED_ID_LENGTH)
               payload.each_with_index do |msg, idx|
                 msg['id'] = "#{idempotent_publish_id}:#{idx}"
               end
--- a/lib/ably/rest/channel.rb
+++ b/lib/ably/rest/channel.rb
@@ -42,7 +42,7 @@ module Ably
       #
       # @param name [String, Array<Ably::Models::Message|Hash>, nil]   The event name of the message to publish, or an Array of [Ably::Model::Message] objects or [Hash] objects with +:name+ and +:data+ pairs
       # @param data [String, ByteArray, nil]   The message payload unless an Array of [Ably::Model::Message] objects passed in the first argument
-      # @param attributes [Hash, nil]   Optional additional message attributes such as :client_id or :connection_id, applied when name attribute is nil or a string
+      # @param attributes [Hash, nil]   Optional additional message attributes such as :extras, :id, :client_id or :connection_id, applied when name attribute is nil or a string
       # @return [Boolean]  true if the message was published, otherwise false
       #
       # @example
@@ -67,6 +67,10 @@ module Ably
         messages = if name.kind_of?(Enumerable)
           name
         else
+          if name.kind_of?(Ably::Models::Message)
+            raise ArgumentError, "name argument does not support single Message objects, only arrays of Message objects"
+          end
+
           name = ensure_utf_8(:name, name, allow_nil: true)
           ensure_supported_payload data
           [{ name: name, data: data }.merge(attributes)]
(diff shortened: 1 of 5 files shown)


mattheworiordan:
Note this does not pass in CI as sandbox does not support idempotent publishing, realtime branch `integration-sparse-serials`

lib/ably/rest/channel.rb
```diff
@@ -58,6 +58,10 @@ def publish(name, data = nil, attributes = {})
         messages = if name.kind_of?(Enumerable)
           name
         else
+          if name.kind_of?(Ably::Models::Message)
```
funkyboy: Maybe this suggestion breaks some requirement in the spec, but I'd be nice if the the lib automatically built an array when just one message is passed as an argument.

spec/acceptance/rest/message_spec.rb
```diff
@@ -91,6 +91,109 @@
...
+        end
+      end
+
+      context 'when ID is included' do
+        context 'with Message object' do
+          let(:message) { Ably::Models::Message.new(id: id, data: data) }
+
+          it 'three REST publishes result in only one message being published' do
+            3.times { channel.publish [message] }
+            expect(channel.history.items.length).to eql(1)
```
funkyboy: Why not testing also the message id here?

lib/ably/rest/channel.rb
```diff
@@ -74,7 +78,12 @@ def publish(name, data = nil, attributes = {})
             unless client.auth.can_assume_client_id?(msg.client_id)
               raise Ably::Exceptions::IncompatibleClientId.new("Cannot publish with client_id '#{msg.client_id}' as it is incompatible with the current configured client_id '#{client.client_id}'")
             end
-          end.as_json
+          end.as_json.tap do |msg|
+            # Mutate the JSON to support idempotent publishing where a Message.id does not exist
+            unless msg['id']
+              msg['id'] = SecureRandom.hex(16).force_encoding('UTF-8')
+            end
```
SimonWoolf: so shouldn't this be conditional on `if client.idempotent_rest_publishing`?

SimonWoolf: also, why not save some bandwidth and use base64 rather than hex, 50% more entropy per character

lib/ably/rest/channel.rb
mattheworiordan: Yeh, well I was thinking of doing that, but it's not defined in the spec elsewhere so wasn't sure I wanted to add a new overloaded method...

lib/ably/rest/channel.rb
mattheworiordan: > so shouldn't this be conditional on client.idempotent_rest_publishing?

Yes.  This change did in fact break some tests, which I never got around to checking.  Good spot, but thankfully the tests broke too as expected!

I've also implemented the `randomId:0` functionality but it gets rejected by realtime:

```
19:49:24.877 [debug] => URL: post https://local-rest.ably.io/channels/e60f1f1eb9b009f5f5399c1c1ce5f950/publish, Headers: Content-Type: "application/json", Accept: "application/json", User-Agent: "Ably Ruby client 1.0.7 (https://www.ably.io)", X-Ably-Version: "1.0", X-Ably-Lib: "ruby-1.0.7", Authorization: "Basic X3RtcF9lRVdWcFEuYzMwUFp3Omtab09MQnpKWXZfUUEyRk0="
19:49:24.878 [debug] => Body: [{"data":"7724cf521aab5a3bd8fd1c18229948b8","id":"is2mDOHSn30fUDFhY/Bnxw==:0"},{"data":"7724cf521aab5a3bd8fd1c18229948b8","id":"is2mDOHSn30fUDFhY/Bnxw==:1"},{"data":"7724cf521aab5a3bd8fd1c18229948b8","id":"is2mDOHSn30fUDFhY/Bnxw==:2"}]
19:49:24.918 [debug] <= Status: 400, Headers: X-Ably-ErrorCode: "40007", X-Ably-ErrorMessage: "Invalid request: multiple messages in a single request with a client-specified id", Content-Type: "application/json", Content-Length: "153", Access-Control-Expose-Headers: "Link,Transfer-Encoding,Content-Length,X-Ably-ErrorCode,X-Ably-ErrorMessage,X-Ably-ServerId,Server", Access-Control-Allow-Origin: "*", Access-Control-Allow-Credentials: "true", Vary: "Origin", Date: "Tue, 24 Jul 2018 18:49:24 GMT", Connection: "keep-alive", Ably-Middleware-Parsed: true
19:49:24.918 [debug] <= Body: {"error"=>{"statusCode"=>400, "code"=>40007, "message"=>"Invalid request: multiple messages in a single request with a client-specified id"}}
```

@SimonWoolf @paddybyers what is wrong with that publish request?

spec/acceptance/rest/message_spec.rb
mattheworiordan: Ok, fixing now.

lib/ably/rest/channel.rb
paddybyers: > what is wrong with that publish request?

> Invalid request: multiple messages in a single request with a client-specified id

According to what's currently implemented based on my proposal (https://github.com/ably/wiki/issues/191#issuecomment-397741807) we only support a single message in a REST publish if the client supplies the `id`.

The final spec is up for discussion.

mattheworiordan:
@paddybyers @funkyboy updated with a64ed80 to reflect finalised spec https://github.com/ably/docs/pull/473.

This does not pass though because batch publishing via REST in realtime is not yet supported.  Will retry once that is enabled.

Issue #164, opened 2018-07-20
Support transient publishes as part of 1.1 spec


mattheworiordan:
See https://github.com/ably/docs/pull/323/files

PR #166, opened 2018-07-22
Transient publishing for #164

mattheworiordan:
This PR adds support for transient publishing as defined in the spec atL:

* https://github.com/ably/docs/pull/323/files
* https://github.com/ably/docs/pull/470

I thought it would be useful to additionally see what's involved in support transient publishing from the proposed `Client#publish` method defined at https://github.com/ably/docs/issues/468.  Interested to hear your views on this.

### DIFF
--- a/lib/ably/realtime/channel.rb
+++ b/lib/ably/realtime/channel.rb
@@ -152,7 +152,7 @@ module Ably
       #   end
       #
       def publish(name, data = nil, attributes = {}, &success_block)
-        if detached? || detaching? || failed?
+        if suspended? || failed?
           error = Ably::Exceptions::ChannelInactive.new("Cannot publish messages on a channel in state #{state}")
           return Ably::Util::SafeDeferrable.new_and_fail_immediately(logger, error)
         end
@@ -372,11 +372,7 @@ module Ably
 
         __queue__.push(*messages)
 
-        if attached?
-          process_queue
-        else
-          attach
-        end
+        process_queue
 
         if messages.count == 1
           # A message is a Deferrable so, if publishing only one message, simply return that Deferrable
@@ -417,7 +413,7 @@ module Ably
 
       # Move messages from Channel Queue into Outgoing Connection Queue
       def process_queue
-        condition = -> { attached? && messages_in_queue? }
+        condition = -> { messages_in_queue? }
         non_blocking_loop_while(condition) do
           send_messages_within_protocol_message __queue__.shift(MAX_PROTOCOL_MESSAGE_BATCH_SIZE)
         end
--- a/lib/ably/realtime/channel.rb
+++ b/lib/ably/realtime/channel.rb
@@ -1,3 +1,5 @@
+require 'ably/realtime/channel/publisher'
+
 module Ably
   module Realtime
     # The Channel class represents a Channel belonging to this application.
@@ -32,6 +34,7 @@ module Ably
       include Ably::Modules::EventMachineHelpers
       include Ably::Modules::AsyncWrapper
       include Ably::Modules::MessageEmitter
+      include Ably::Realtime::Channel::Publisher
       extend Ably::Modules::Enum
 
       # ChannelState
@@ -59,7 +62,7 @@ module Ably
       # Max number of messages to bundle in a single ProtocolMessage
       MAX_PROTOCOL_MESSAGE_BATCH_SIZE = 50
 
-      # Ably client associated with this channel
+      # {Ably::Realtime::Client} associated with this channel
       # @return [Ably::Realtime::Client]
       # @api private
       attr_reader :client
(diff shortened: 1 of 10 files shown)


lib/ably/realtime/channel.rb
```diff
@@ -146,13 +146,13 @@ def initialize(client, name, channel_options = {})
...
       def publish(name, data = nil, attributes = {}, &success_block)
-        if detached? || detaching? || failed?
+        if suspended? || failed?
           error = Ably::Exceptions::ChannelInactive.new("Cannot publish messages on a channel in state #{state}")
           return Ably::Util::SafeDeferrable.new_and_fail_immediately(logger, error)
         end
 
         if !connection.can_publish_messages?
-          error = Ably::Exceptions::MessageQueueingDisabled.new("Message cannot be published. Client is configured to disallow queueing of messages and connection is currently #{connection.state}")
+          error = Ably::Exceptions::MessageQueueingDisabled.new("Message cannot be published. Client is not allowed to queue of messages when connection is in state #{connection.state}")
```
SimonWoolf: "to queue of messages" -> "to queue messages"

lib/ably/realtime/channel.rb
```diff
@@ -164,7 +164,12 @@ def publish(name, data = nil, attributes = {}, &success_block)
           [{ name: name, data: data }.merge(attributes)]
         end
 
-        queue_messages(messages).tap do |deferrable|
+        if messages.length > Realtime::Connection::MAX_PROTOCOL_MESSAGE_BATCH_SIZE
+          error = Ably::Exceptions::InvalidRequest.new("It is not possible to publish more than #{Realtime::Connection::MAX_PROTOCOL_MESSAGE_BATCH_SIZE} messages with a single publish request.")
+          return Ably::Util::SafeDeferrable.new_and_fail_immediately(logger, error)
+        end
```
SimonWoolf: Where'd this MAX_PROTOCOL_MESSAGE_BATCH_SIZE requirement come from? Don't remember any discussion on this, and it's not in the spec afaict

lib/ably/realtime/channel/channel_manager.rb
```diff
@@ -111,16 +111,12 @@ def fail_messages_awaiting_ack(error, options = {})
         end
       end
 
-      # When a channel becomes detached, suspended or failed,
+      # When a channel becomes suspended or failed,
       # all queued messages should be failed immediately as we don't queue in
       # any of those states
       def fail_queued_messages(error)
         error = Ably::Exceptions::MessageDeliveryFailed.new("Queued messages on channel '#{channel.name}' in state '#{channel.state}' will never be delivered") unless error
         fail_messages_in_queue connection.__outgoing_message_queue__, error
```
SimonWoolf: > `fail_messages_in_queue connection.__outgoing_message_queue__, error`

Not something you added in this pr, but: am I misunderstanding something, or does this fail every message in the connection queue if any channel becomes suspended or failed? That's not right, surely

lib/ably/realtime/client/incoming_message_dispatcher.rb
```diff
@@ -102,7 +102,11 @@ def dispatch_protocol_message(*args)
               if channel.attached?
                 channel.manager.duplicate_attached_received protocol_message
               else
-                channel.transition_state_machine :attached, reason: protocol_message.error, resumed: protocol_message.has_channel_resumed_flag?, protocol_message: protocol_message
+                if channel.failed?
+                  logger.warn "Ably::Realtime::Client::IncomingMessageDispatcher - Received an ATTACHED protocol message for FAILED channel #{channel.name}. Ignoring ATTACHED message"
```
SimonWoolf: Is there a spec for this? Having the client deliberately ignore what the server is telling it the channel state is is a bit of a break from our usual model (of the server deciding and telling the client), no?

lib/ably/realtime/channel.rb
mattheworiordan: Thanks, good spot.

lib/ably/realtime/channel.rb
mattheworiordan: See https://github.com/ably/ably-ruby/pull/166/commits/fd267e6599e47f25eb81551924f5322a321ce386.  I mentioned in that commit that this limit existed before, so I have kept it as it was.  @paddybyers I believe we have a max `ProtocolMessage` size, but do we have / want any limit on how many messages can exist within a single `ProtocolMessage`?

lib/ably/realtime/channel/channel_manager.rb
mattheworiordan: No, that is not correct, see https://github.com/ably/ably-ruby/pull/166/files/f7a524e00e64a47cf5c9d974dbcf621f5b82307a#diff-b8c5e1edac0c00c8db3f3e6ebcc8e749R125, only the messages for that channel.   However, based on what we agree on https://github.com/ably/docs/pull/470, I'll most likely need to modify this...

lib/ably/realtime/client/incoming_message_dispatcher.rb
mattheworiordan: This wasn't a change of behaviour.  This was raising an exception before because a channel is not allowed to transition from `failed` directly to `attached`.  See https://github.com/ably/ably-ruby/blob/f7a524e00e64a47cf5c9d974dbcf621f5b82307a/lib/ably/realtime/channel/channel_state_machine.rb#L29.  I perhaps should investigate more under what conditions this has happened (I saw this exception bubble up in one of the tests), but I'd prefer to create a separate issue as it's not related to this PR, this change just made the library handle the problem better.

mattheworiordan:
https://github.com/ably/docs/pull/470 is merged, so I need to implement this change before this is ready for review again.

PR #169, opened 2018-08-11
Parallel tests

mattheworiordan:
Tests are taking a very long time now as there are so many.  Adding the recent push tests pushed the test suite time over 1 hour.

This PR aims to fix that by parallelising the tests over 6 processes with Travis, and more on our internal CI box.

P.S. My commit naming convention is a shambles, I'll fix when I rebase and merge :)


mattheworiordan:
Results are good so far.  I've managed to get the Travis CI tests from 40+ minutes down to less than 10 minutes 👍.

lib/ably/rest/client.rb
```diff
@@ -140,18 +145,19 @@ def initialize(options)
...
-        @protocol         = options.delete(:protocol) || :msgpack
-        @debug_http       = options.delete(:debug_http)
-        @log_level        = options.delete(:log_level) || ::Logger::WARN
-        @custom_logger    = options.delete(:logger)
-        @custom_host      = options.delete(:rest_host)
-        @custom_port      = options.delete(:port)
-        @custom_tls_port  = options.delete(:tls_port)
-        @add_request_ids  = options.delete(:add_request_ids)
+        @realtime_client     = options.delete(:realtime_client)
+        @tls                 = options.delete(:tls) == false ? false : true
```
SimonWoolf: `!(options.delete(:tls) == false)`

mattheworiordan: Sure, we could do lots of Ruby golf throughout, but I prefer to have the code as readable as possible. So it's a style preference of course, but I think a ternary expression is easier to comprehend.  BTW. This code was not changed, it was just indented.

mattheworiordan:
Seems Github is incorrectly reporting this as failed, yet it passed at https://travis-ci.org/ably/ably-ruby/builds/415361140, with test duration coming in at an average of 8 minutes 🍾

Issue #163, opened 2018-06-21
my-members presenceMap requirement change for 1.1


SimonWoolf:
See https://github.com/ably/docs/pull/446

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-464) by [Unito](https://www.unito.io/learn-more)

SimonWoolf:
> the expected scenario this is going to happen is when a client enters a channel, becomes disconnected for for more than 15 seconds, but less than 2 minutes, then resumes their connection, and now receives a synthesized leave because they've been disconnected for more than 15 seconds right?

Correct

> I assume just ignoring the synthesized leave event completely (in regards to PresenceMap, we'll still emit the event) and waiting for the new enter event from Realtime (which too can effectively be ignored, although I won't ignore it as I'll use the latest timestamp etc.) is the right thing to do.

So I guess that should work, because even if you ignore the synthesized leave, if realtime doesn't re-enter you, your members will still be missing from the sync set and will be marked absent from the main presence set that way, so the client lib will notice and re-enter them.

But unless I'm missing something, AFAICS you also don't actually need to do that. The synthesized leave already won't affect the local members map with ably-ruby at the moment, because your `member_set_delete` only touches the local members hash if the channel is `in_sync`. A synthesized leave will be received between the ATTACHED and the SYNC, so the channel will not be in_sync.  (And if it ever comes after the SYNC, well, by then it'll be too late, the lib will already have re-entered its absent members. So it should work either way).

So I haven't tried it, but looking at the code I think ably-ruby should already do the right thing in this situation, as a side effect of the way you've done it.

(Which is nice as TBH I'm not a big fan of the idea of ignoring presence messages (wrt the main map) anyway. E.g what if at some point we stop sending a SYNC after a resume (nothing in the spec promises we'll always do that), realtime doesn't re-enter you, and then your re-enter fails with max-member-count-exceeded? the map will be left in an incorrect state. OK, that sounds a bit farfetched, but you know what I mean, ignoring it feels wrong)

> Have you got a test for this scenario?

Not in the ably-js repo, but I've a comprehensive test [in the realtime repo](https://github.com/ably/realtime/blob/develop/test/connection/presenceresume.js#L36-L51) that uses ably-js -- having it there means I can use the farm + status api to force the connection to resume or recover to a different frontend (as if the previous one had downscaled) to properly test the scenario where the client lib will need to auto-re-enter.

PR #173, opened 2018-11-13
V1.1 release

mattheworiordan:
@paddybyers @SimonWoolf I think Ruby is ready for a 1.1 release, so running this through CI.

Some PRs I would like to incorporate:

* [Transient publishing](https://github.com/ably/ably-ruby/pull/166)
* [Idempotent publishing](https://github.com/ably/ably-ruby/pull/165).  I believe that whilst support for the full spec is not yet fully rolled out in production, it's safe to merge this.

I will rebase those PRs on this branch now.


paddybyers:
According to https://github.com/ably/docs/pull/473/files#diff-11900b395df266bc2bbfc1d11bdfc7a0R1232, the behaviour of idempotent publishing also needs to change with the switch to 1.1.

SimonWoolf:
@mattheworiordan would be nice to confirm my analysis in https://github.com/ably/ably-ruby/issues/163#issuecomment-407015794 is right (if so you don't need to make any further changes for 1.1 re presence)

mattheworiordan:
>  would be nice to confirm my analysis in #163 (comment) is right (if so you don't need to make any further changes for 1.1 re presence)

Ok, will prioritise

mattheworiordan:
>  the behaviour of idempotent publishing also needs to change with the switch to 1.1.

@paddybyers OH. I implemented 1.2 in the Ruby lib :(  

That in theory is a bit of a blocker no?  Thoughts?

mattheworiordan:
@paddybyers @SimonWoolf I believe this PR is ready to be merged and released as 1.1, barring any minor fixes in broken tests (such as idempotent tests which won't pass). See the "Ably Realtime Unified Spec 1.1" now, up to 88% coverage now.

screenshot 2018-11-28 at 00 45 46

PR #196, opened 2020-02-21
Use default fallback hosts when custom environment set

mattheworiordan:
When environment is set in the client SDKs, we can now rely on default fallback hosts as follows.

Fallbacks are ENV-[a-e]-fallback.ably-realtime.com. For example:

ENV-a-fallback.ably-realtime.com
ENV-b-fallback.ably-realtime.com
ENV-c-fallback.ably-realtime.com
ENV-d-fallback.ably-realtime.com
ENV-e-fallback.ably-realtime.com

Note that this change:

- It breaks a lot of tests against sandbox that assumed no fallback hosts would be used
- We need to ensure that custom hosts or custom ports result do not use default environment fallback hosts.  We should only use when only the custom environment is set
- Some spec updates are needed in the spec relating to the use of fallback hosts

See related issue https://github.com/ably/docs/issues/818 (no spec yet exists)


lmars:
Just to note, we should not release this until we've run a DNS update to ensure all the fallbacks exist.

lmars:
DNS updates have all been run so this is no longer blocked on that.

In response to points in the description:

> It breaks a lot of tests against sandbox that assumed no fallback hosts would be used

Sandbox now has fallback hosts configured, in case that was the reason for the failures?

> Some spec updates are needed in the spec relating to the use of fallback hosts

The spec was updated in https://github.com/ably/docs/pull/965.

SimonWoolf:
@TheSmartnik @QuintinWillison This PR was made before the spec https://github.com/ably/docs/pull/965/files was finalised. It may all be fine, but should go through that spec and check that this PR implements the final spec correctly

jamienewcomb:
@TheSmartnik did you managed to look at Simons comment 
https://github.com/ably/ably-ruby/pull/196#issuecomment-745384245

If you are happy could you approve and we will get this merged

Thanks

TheSmartnik:
@jamienewcomb sorry, I've missed the comment. I'll double-check it later today and reply

lmars:
@TheSmartnik it would be great to get this merged, did you get chance to review Simon's comment?

PR #207, opened 2020-10-22
Update specs to match backend response

TheSmartnik:


spec/acceptance/realtime/auth_spec.rb
```diff
@@ -1058,7 +1058,7 @@ def disconnect_transport(connection)
 
           it 'disconnected includes and invalid signature message' do
             client.connection.once(:disconnected) do |state_change|
-              expect(state_change.reason.message.match(/invalid signature/i)).to_not be_nil
+              expect(state_change.reason.message.match(/signature verification failed/i)).to_not be_nil
```
lmars, line 1061: I think this would be better as:

```ruby
expect(state_change.reason.message).to match(/signature verification failed/i)
```

lmars (approved):
A minor comment, but LGTM 👍

spec/acceptance/realtime/auth_spec.rb
TheSmartnik: @lmars right, thanks

TheSmartnik: @lmars I've checked other specs and they are all in this format. I would rather refactor it separately. Also, specs are very flaky as of now and small change will result in multiple rebuilds

lmars: 👍

Issue #208, opened 2020-12-07
Replace Travis with GitHub workflows


QuintinWillison:
Discuss with @QuintinWillison when you come to this as our adoption of this tech is evolving rapidly at the moment.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-541) by [Unito](https://www.unito.io/learn-more)

owenpearson:
Closing as this work has already been done

PR #214, opened 2021-01-30
Update Travis CI versions

TheSmartnik:
I know we're planning to Github workflows #208 but before that, following the discussion on Slack I've updated ruby to the current versions, removed redundant and EOL ones. I'll speed up our test suit a bit


owenpearson (approved):
LGTM, thanks.

Issue #199, opened 2020-06-12
statesman dependency very outdated


darkhelmet:
We use Ably in our Rails app and it's great.

We would like to use statesman but the Ably gem limits us to 1.0.0, which is a problem.

Is it possible to get that dependency upgraded so we can use a modern version of statesman?

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/SDK-479) by [Unito](https://www.unito.io/learn-more)

QuintinWillison:
Hello @darkhelmet ... thanks for flagging this. We're taking a look.
([internal question](https://ably-real-time.slack.com/archives/C030C5YLY/p1592219605392900))

mattheworiordan:
@darkhelmet in principle of course we're happy to flex the requirements, we just need to see what impact it has on the Ably library itself (i.e. it looks like statesman is now on 7.2.0, and we're on 1.0.0, which in theory means 6 breaking changes).   In addition, if we bump to 7.*, then sadly we're also going to break anyone else who depends on a lower version of statesmen and can't update.  Sadly this is the major shortfall in Ruby and Gems (support for conflicting dependencies).

Two quick questions anyway so that we can find a solution:

* Which version of Statesman are you using?
* Are you intentionally using the Realtime SDK as opposed to just the REST SDK?  See https://support.ably.io/support/solutions/articles/3000060319-should-i-use-the-rest-or-realtime-library to understand the difference, and https://github.com/ably/docs/issues/883 which is our recognition that this is a problem we may need to fix our docs.

darkhelmet:
We haven't started using statesman yet, but want to for an upcoming set of features. I had seen the gem a while ago, had it bookmarked, and saw it in our Gemfile.lock and did the "oh yeah, that would be perfect" reaction. Given our desire to keep things up to date, I'd want to start using the latest version possible for the gem.

We currently only use the REST part of the Ruby SDK, though we are looking at ripping out all our ActionCable usage for Ably and I'm not sure what exactly that will require us to use, so we might need to get into the Realtime SDK in the future.

PR #217, opened 2021-02-02
Upgrade statesman

TheSmartnik:
Since all changes to statesman are related to ActiveRecord, we can easily update the version of statesman without any changes. This Mr is based https://github.com/ably/ably-ruby/pull/214 since statesman doesn't support EOL versions of Ruby.

Closes: https://github.com/ably/ably-ruby/issues/199


owenpearson (approved):
LGTM, thanks.

PR #227, opened 2021-03-06
Add ChannelProperties (RTL15)

TheSmartnik:


spec/acceptance/realtime/channel_spec.rb
```diff
@@ -83,6 +83,16 @@ def disconnect_transport
           end
         end
 
+        it 'sets attach_serial property after the attachment (#RTL15a)' do
+          expect(channel.properties.attach_serial).to be_nil
+
+          channel.attach
+          channel.on(:attached) do
+            expect(channel.properties.attach_serial).to_not be_nil
```
owenpearson, line 91: Can we add a test (or add something to this test) to make sure that when we call `channel.history` it sends the serial as part of the options? I think that might be a more important thing to check rather than just whether the serial exists.

owenpearson (approved):
Just one small comment, but overall looks good :+1:

spec/acceptance/realtime/channel_spec.rb
TheSmartnik: @owenpearson there are few specs for `untilAttach` which covers pretty much what you're asking
https://github.com/ably/ably-ruby/blob/main/spec/acceptance/realtime/channel_history_spec.rb#L167-L219

Does it count?

owenpearson: Oh nice, yeah that covers it.

PR #231, opened 2021-04-12
Validate that members presenceMap does not change on synthesized leave

TheSmartnik:
Resolves https://github.com/ably/ably-ruby/issues/163


owenpearson (approved):
LGTM, thanks

PR #237, opened 2021-04-22
Upgrade statesman to ~> 8.0

darkhelmet:
statesman is on 8.x now which fixes a Rails 6.1 related deprecation.

In [v8.0.0 on Jan 6](https://github.com/gocardless/statesman/blob/master/CHANGELOG.md#v800-6th-january-2021) they fixed a deprecation, specifically [Use AR Arel table to type cast booleans in order to avoid deprecation warning #421](https://github.com/gocardless/statesman/pull/421)

Without this you get a bunch of deprecation messages, and obviously it'll stop working in Rails 6.2 as the deprecation points out.

This change allows the use of the 8.x line of stateman with ably so that we can upgrade to versions with the fix.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        ruby: [ '2.5', '2.6', '2.7' ]
+        ruby: [ '2.5', '2.6', '2.7', '3.0' ]
         protocol: [ 'json', 'msgpack' ]
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -10,7 +10,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        ruby: [ '2.5', '2.6', '2.7', '3.0' ]
+        ruby: [ '2.5', '2.6', '2.7' ]
         protocol: [ 'json', 'msgpack' ]
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


darkhelmet:
Looks like the ruby 3 specs passed but failed submitting to coveralls.

QuintinWillison:
Thanks, @darkhelmet. 😁 I'll re-run those jobs.

Meanwhile, can you give any more detail around what the Rails 6.1 related deprecation was, how it was affecting you and how this update fixes it, even if just a pointer to somewhere else? I took a quick look at the [statesman changelog](https://github.com/gocardless/statesman/blob/master/CHANGELOG.md) and couldn't see anything obvious.

darkhelmet:
@QuintinWillison Updated the summary with some links and more details.

lukaszsliwa (approved):

darkhelmet:
Woo! @lukaszsliwa Do you need to authorize the workflow run to get it fully green?

lukaszsliwa:
> Woo! @lukaszsliwa Do you need to authorize the workflow run to get it fully green?

There is a problem with `eventmachine not initialized: evma_signal_loopbreak` in the workflow. Anyway, I find out that I don't have merge permission...
I have researched there are more spec with this, so probably it would be better to resolve it in the main branch. However, I remember that this problem with `evma_signal_loopbreak` from Eventmachine, comes back from time to time like boomerang.

lmars:
It looks like the two Ruby 3 builds are failing with:

```
/home/runner/work/ably-ruby/ably-ruby/spec/acceptance/rest/client_spec.rb:3:in `require': cannot load such file -- webrick (LoadError)
```

Whilst that is something that certainly needs fixing (I believe webrick is no longer bundled with Ruby 3), I do wonder if it's worth dealing with Ruby 3 support in a separate PR? Or is Ruby 3 support somehow required for the statesman upgrade?

darkhelmet:
We're using this in Ruby 3 just fine, so I figured I'd turn on 3.0 in the matrix. I imagine it's just a spec failure, but yeah, a different PR for that upgrade.

lmars (approved):

lmars:
@darkhelmet Thanks!

PR #248, opened 2021-05-24
Implement RSC7d (Ably-Agent header)

lukaszsliwa:
#230 

https://github.com/ably/docs/pull/1034

- Removed `X-Ably-Lib` header param
- Added `Ably-Agent` header param to the REST client headers
- overwrite agent for example `Ably::Rest::Client.new(agent: 'ably-ruby/1.1.5 ruby/1.9.3')`
- Added `agent` parameter to the realtime client connection auth_deferrable params
- Removed `Ably.lib_variant`
- Added `Ably::AGENT` in `lib/ably/agent.rb`
- Fixed specs
- Updated `ably-ruby-rest` https://github.com/ably/ably-ruby-rest/pull/17

### DIFF
--- a/lib/ably.rb
+++ b/lib/ably.rb
@@ -1,6 +1,7 @@
 require 'addressable/uri'
 
 require 'ably/version'
+require 'ably/agent'
 
 %w(modules util).each do |namespace|
   Dir.glob(File.expand_path("ably/#{namespace}/*.rb", File.dirname(__FILE__))).sort.each do |file|
--- a/lib/ably/agent.rb
+++ b/lib/ably/agent.rb
@@ -1,0 +1,4 @@
+module Ably
+  # Default Ably Agent
+  AGENT = "ably-ruby/#{Ably::VERSION}"
+end
--- a/lib/ably/agent.rb
+++ b/lib/ably/agent.rb
@@ -1,4 +1,3 @@
 module Ably
-  # Default Ably Agent
-  AGENT = "ably-ruby/#{Ably::VERSION}"
+  AGENT = "ably-ruby/#{Ably::VERSION} ruby/#{RUBY_VERSION}"
 end
--- a/lib/ably/realtime/connection.rb
+++ b/lib/ably/realtime/connection.rb
@@ -434,7 +434,7 @@ module Ably
                 'format' =>     client.protocol,
                 'echo' =>       client.echo_messages,
                 'v' =>          Ably::PROTOCOL_VERSION,
-                'lib' =>        client.rest_client.lib_version_id,
+                'agent' =>      client.rest_client.agent
               )
 
               # Use native websocket heartbeats if possible, but allow Ably protocol heartbeats
--- a/lib/ably/rest/client.rb
+++ b/lib/ably/rest/client.rb
@@ -42,9 +42,6 @@ module Ably
         Faraday::ClientError
       end
 
-      # Default Ably Agent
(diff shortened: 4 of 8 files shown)


owenpearson (approved):
Looks really good, thanks 🙂 

My only question is: how will this work with our `ably-ruby-rest` library? This is our synchronous REST-only ruby SDK which pulls in `ably-ruby` as a submodule. I realise I haven't mentioned in #230, but it would be great if we could have some way of identifying whether someone is using the REST-only library or not. I'm not sure whether we would need to update this PR to accommodate that or whether you could leave this as it is and make just make a separate PR to the `ably-ruby-rest` repo. I'm happy for you to decide what's best.

lukaszsliwa:
> Looks really good, thanks 🙂
> 
> My only question is: how will this work with our `ably-ruby-rest` library? This is our synchronous REST-only ruby SDK which pulls in `ably-ruby` as a submodule. I realise I haven't mentioned in #230, but it would be great if we could have some way of identifying whether someone is using the REST-only library or not. I'm not sure whether we would need to update this PR to accommodate that or whether you could leave this as it is and make just make a separate PR to the `ably-ruby-rest` repo. I'm happy for you to decide what's best.


Thanks. To change the `Ably-Agent` header for the ably-ruby-rest, we need to change this file:
or add `lib/agent.rb` like the following:

```
module Ably
  AGENT = "ably-ruby-rest/#{Ably::VERSION}"
end

```
and ignore loading `lib/agent.rb` from the modules.

owenpearson:
@lukaszsliwa I've checked with others on the team and we would prefer for the `Ably-Agent` for `ably-ruby-rest` to look like this:
```
AblyAgent: ably-ruby/1.2.6 ruby/2.6.6 ably-ruby-rest
```
So we would still have the `ably-ruby` identifier for both libraries, and have the `ably-ruby-rest` identifier without a version number (ably-ruby and ably-ruby-rest should always have the same version). Let me know if there are any problems with implementing it in this way?

lukaszsliwa:
> ably-ruby/1.2.6 ruby/2.6.6 ably-ruby-rest



> @lukaszsliwa I've checked with others on the team and we would prefer for the `Ably-Agent` for `ably-ruby-rest` to look like this:
> 
> ```
> AblyAgent: ably-ruby/1.2.6 ruby/2.6.6 ably-ruby-rest
> ```
> 
> So we would still have the `ably-ruby` identifier for both libraries, and have the `ably-ruby-rest` identifier without a version number (ably-ruby and ably-ruby-rest should always have the same version). Let me know if there are any problems with implementing it in this way?

No problems. I have already pushed a change. https://github.com/ably/ably-ruby/pull/248/commits/902d5d5901f03c60f2a3286d1b56b515c5f174b1


For the ably-ruby-rest we need to open another pull-request in https://github.com/ably/ably-ruby-rest

```ruby
module Ably
  AGENT = "ably-ruby/#{Ably::VERSION} ruby/#{RUBY_VERSION} ably-ruby-rest"
end
```

owenpearson:
@lukaszsliwa would you mind creating this PR in `ably-ruby-rest`? You should have received an invite for access to that repo 🙂

lukaszsliwa:
> @lukaszsliwa would you mind creating this PR in `ably-ruby-rest`? You should have received an invite for access to that repo 🙂

Sure, I will update ably-ruby-rest too.

lukaszsliwa:
@owenpearson 

I have opened pull-request https://github.com/ably/ably-ruby-rest/pull/17

PR #257, opened 2021-07-12
Add a spec for #RTN11c

TheSmartnik:
As a part of https://github.com/ably/ably-ruby/issues/244

### DIFF
--- a/spec/acceptance/realtime/connection_spec.rb
+++ b/spec/acceptance/realtime/connection_spec.rb
@@ -482,6 +482,89 @@ describe Ably::Realtime::Connection, :event_machine do
         end
       end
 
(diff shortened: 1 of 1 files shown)


lukaszsliwa:
`DISCONNECTED` state confirmation is missing. `SUSPENDED` connection status works. I think it would be good to divide these tests in contexts base on the connection status.

```
(RTN11c) If the state is DISCONNECTED or SUSPENDED, skips the ongoing wait in the retry process described in RTN14d and RTN14e, so that the next reconnection attempt happens immediately.
```

TheSmartnik:
@lukaszsliwa done

lukaszsliwa (approved):
Looks good to me. Thanks @TheSmartnik

owenpearson (approved):
LGTM, thanks

PR #258, opened 2021-07-13
Try to fix flaky specs

TheSmartnik:
1. This mr changes specs to avoid race conditions
2. Removes specs that no longer have a spec
3. Skips the ones that are constantly failing due to changes in sandbox env


owenpearson (commented):
The `xit` and adding timers to avoid race conditions look sensible to me. The only bit I'm not sure about is "Removes specs that no longer have a spec"; could you give a bit more explanation about what this means and why we should remove these specs?

TheSmartnik:
@owenpearson `RTP5c1` and `RTP5c2` are no longer exist https://docs.ably.io/client-lib-development-guide/features/#RTP5
I'd try to fix them but the behavior they testing are not part of the doc anyore

owenpearson (approved):
LGTM

PR #309, opened 2021-08-02
Using xit() instead of pending() inside it() {} block

lukaszsliwa:
`pending` inside the `it() { ... }` block will still execute callbacks, for example `before` callback.

Marked one spec with `xit()` and commented `pending()`


TheSmartnik:
> pending inside the it() { ... } block will still execute callbacks, for example before callback

Good idea. Thanks for the fix!

PR #322, opened 2021-08-30
Add support for DataTypes ProtocolMessage: has_attach_resume_flag?

### LINKED ISSUE #274: Add support for DataTypes ProtocolMessage TR3f, TR4i, q
Add support for the following requirements: TR3f, TR4i, q

As stated at https://docs.ably.io/client-lib-development-guide/features/#TR3f https://docs.ably.io/client-lib-development-guide/features/#TR4i Message

see details https://github.com/ably/docs/blob/main/content/client-lib-development-guide/versions/features-1-1__1-2.diff

Acceptance Criteria: 

Implement the support for the requirements above

Update the spreadsheet

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1076) by [Unito](https://www.unito.io)

lukaszsliwa:
TR3f

Resolves #274 

Added `ProtocolMessage#has_attach_resume_flag?`

### DIFF
--- a/lib/ably/models/protocol_message.rb
+++ b/lib/ably/models/protocol_message.rb
@@ -217,6 +217,11 @@ module Ably::Models
     end
 
     # @api private
+    def has_attach_resume_flag?
+      flags && 32 == 32 # 2^5
+    end
+
+    # @api private
     def has_attach_presence_flag?
       flags & 65536 == 65536 # 2^16
     end
--- a/lib/ably/models/protocol_message.rb
+++ b/lib/ably/models/protocol_message.rb
@@ -218,7 +218,7 @@ module Ably::Models
 
     # @api private
     def has_attach_resume_flag?
-      flags && 32 == 32 # 2^5
+      flags & 32 == 32 # 2^5
     end
 
     # @api private
--- a/spec/unit/models/protocol_message_spec.rb
+++ b/spec/unit/models/protocol_message_spec.rb
@@ -176,6 +176,14 @@ describe Ably::Models::ProtocolMessage do
         end
       end
 
+      context 'when attach resumed flag' do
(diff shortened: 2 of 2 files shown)


lib/ably/models/protocol_message.rb
```diff
@@ -216,6 +216,11 @@ def has_transient_flag?
       flags & 16 == 16 # 2^4
     end
 
+    # @api private
+    def has_attach_resume_flag?
+      flags && 32 == 32 # 2^5
```
owenpearson, line 221: This should be a single `&` (bitwise 'and' operator). At the moment it will return true whenever `flags` is truthy.

spec/unit/models/protocol_message_spec.rb
```diff
@@ -176,6 +176,14 @@ def new_protocol_message(options)
         end
       end
 
+      context 'when attach resumed flag' do
```
owenpearson, line 179: Might be worth testing the negative case as well (if you create a `protocol_message` without the `attach_resume` flag then `has_attach_resume_flag?` returns `false`). A test like that would have made the accidental `&&` obvious.

owenpearson (changes requested):

lib/ably/models/protocol_message.rb
lukaszsliwa: Fixed. Thanks, good catch.

spec/unit/models/protocol_message_spec.rb
lukaszsliwa: Fixed. Added two tests. One test with `flags = 0`:
`0 && 32 == 32` is `false`
`0 & 32 == 32` is `true`

owenpearson (approved):

PR #324, opened 2021-09-13
When connection disconnectes and can't renew token it fails (RTN15h1)

TheSmartnik:
As part of https://github.com/ably/ably-ruby/issues/244

RTN15h2 would be a different pr since it requires a lot more time


owenpearson (approved):
LGTM, thanks

lukaszsliwa (approved):

PR #334, opened 2021-11-08
Update RSC7

TheSmartnik:
Fixes https://ably.atlassian.net/browse/SDK-1084

Well, everything turned out to be already working :D So, I've just updated the specs

### DIFF
--- a/lib/ably/version.rb
+++ b/lib/ably/version.rb
@@ -1,6 +1,6 @@
 module Ably
   VERSION = '1.1.8'
-  PROTOCOL_VERSION = '1.1'
+  PROTOCOL_VERSION = '1.2'
 
   # @api private
   def self.major_minor_version_numeric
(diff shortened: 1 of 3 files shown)


TheSmartnik:
Oh, it also fixes https://ably.atlassian.net/browse/SDK-1085

Three lines, two tickets

lib/ably/version.rb
```diff
@@ -1,6 +1,6 @@
 module Ably
   VERSION = '1.1.7'
-  PROTOCOL_VERSION = '1.1'
+  PROTOCOL_VERSION = '1.2'
```
owenpearson, line 3: As discussed at the previous standup we're not at API version 1.2 yet so I don't agree with this change.

owenpearson (changes requested):

lukaszsliwa (approved):

owenpearson (approved):

Issue #281, opened 2021-08-02
Write spec tests for RSA4b, b1, c, RSA16 (Authentication)


AndyNicks:
Write spec tests for : RSA4b, b1, c, RSA16
ensure they pass and that the affected API conforms to the 1.2 API design

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1069) by [Unito](https://www.unito.io)

PR #338, opened 2021-12-20
Add support for RSA4b, b1, c, RSA16 (Authentication)

### LINKED ISSUE #268: Add support for RSA4b, b1, c, RSA16 (Authentication)
Add support for the following requirements: RSA4b, b1, c, RSA16

As stated at https://docs.ably.io/client-lib-development-guide/features/#RSA4b https://docs.ably.io/client-lib-development-guide/features/#RSA4c https://docs.ably.io/client-lib-development-guide/features/#RSA16

see details https://github.com/ably/docs/blob/main/content/client-lib-development-guide/versions/features-1-1__1-2.diff

Acceptance Criteria: 

Implement the support for the requirements above

Update the spreadsheet

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1082) by [Unito](https://www.unito.io)

lukaszsliwa:
Resolves #268  #281 

- [x] Using current time when checking token expiration (updated TokenDetails#expired? method by adding attributes and `from` option) RSA4b, RSA4b1
- [x] RSA16 is already implemented and tested as `current_token_details`

### DIFF
--- a/lib/ably/auth.rb
+++ b/lib/ably/auth.rb
@@ -510,11 +510,11 @@ module Ably
     end
 
     def authorize_when_necessary
-      if current_token_details && !current_token_details.expired?
+      if current_token_details && !current_token_details.expired?(from: current_time)
         return current_token_details
-      else
-        authorize
       end
+
+      authorize
     end
 
     # Returns the current device clock time unless the
--- a/lib/ably/models/token_details.rb
+++ b/lib/ably/models/token_details.rb
@@ -95,10 +95,15 @@ module Ably::Models
     # Returns true if token is expired or about to expire
     # For tokens that have not got an explicit expires attribute expired? will always return true
     #
+    # @param attributes [Hash]
+    # @option attributes [Time] :from   Sets a current time from which token expires
+    #
     # @return [Boolean]
-    def expired?
+    def expired?(attributes = {})
       return false if !expires
-      expires < Time.now + TOKEN_EXPIRY_BUFFER
+
+      from = attributes[:from] || Time.now
+      expires < from + TOKEN_EXPIRY_BUFFER
     end
 
     # True if the TokenDetails was created from an opaque string i.e. no metadata exists for this token
--- a/spec/unit/models/token_details_spec.rb
+++ b/spec/unit/models/token_details_spec.rb
@@ -80,6 +80,20 @@ describe Ably::Models::TokenDetails do
           expect(subject.expired?).to eql(false)
         end
       end
+
+      context 'with :from attribute' do
+        subject { Ably::Models::TokenDetails.new(expires: expire_time) }
+
+        let(:server_offset_time) { 2 * 60 * 60 } # 2 hours
+
+        it 'is false' do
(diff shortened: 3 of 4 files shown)


owenpearson (approved):
LGTM

lib/ably/auth.rb
```diff
@@ -510,11 +510,11 @@ def token_option
     end
 
     def authorize_when_necessary
-      if current_token_details && !current_token_details.expired?
+      if current_token_details && !current_token_details.expired?(from: current_time)
```
TheSmartnik, line 513: Don't we need to double-check here that `@server_time_offset` is present to make sure that current_time is based on server time?

lib/ably/models/token_details.rb
```diff
@@ -95,10 +95,15 @@ def client_id
     # Returns true if token is expired or about to expire
     # For tokens that have not got an explicit expires attribute expired? will always return true
     #
+    # @param attributes [Hash]
+    # @option attributes [Time] :from   Sets a current time from which token expires
+    #
     # @return [Boolean]
-    def expired?
+    def expired?(attributes = {})
```
TheSmartnik, line 102: Why not use the default option?
```
def expired?(from: Time.now)
```

spec/unit/models/token_details_spec.rb
```diff
@@ -80,6 +80,20 @@
           expect(subject.expired?).to eql(false)
         end
       end
+
+      context 'with :from attribute' do
+        subject { Ably::Models::TokenDetails.new(expires: expire_time) }
```
TheSmartnik, line 85: That's a little picky, but I'd suggest naming a subject as it improves readability 
```
subject(:token_detail) { .. }
```

lib/ably/models/token_details.rb
lukaszsliwa: Such syntax is available since ruby `2.x` but we support `1.9.3+` https://github.com/ably/ably-ruby#supported-platforms

```markdown
This SDK supports Ruby 1.9.3+. For eventmachine and Ruby 3.0 note...
```

I don't know if anyone use this lib on the machine with ruby `1.9.x`

spec/unit/models/token_details_spec.rb
lukaszsliwa: I can change it but it does not matter for `123` occurrences of other "subject"s in +39 files.

```bash
find . -name "*_spec.rb" -exec grep -H "subject {" {} \; | wc -l   
     123
```

lib/ably/auth.rb
lukaszsliwa: I have checked local build. It always calls for `ensure_current_time_is_based_on_server_time` if `query_time` and requesting for a token.

QuintinWillison:
An approval from @TheSmartnik is going to carry much more value than one from me, so I'm going to remove my review request. There is a lot of background context to this change and the spec points referenced are borderline cryptic, so I would prefer to trust in those who are more actively working on this SDK to make the call as to whether this is ready to be merged.

TheSmartnik (approved):

PR #340, opened 2022-01-25
RSL1a, RSL1b

### LINKED ISSUE #283: Write spec tests for RSL1a, b, h, k1, k2, l, l1 (Channels)
Write spec tests for : RSL1a, b, h, k1, k2, l, l1 ensure they pass and that the affected API conforms to the 1.2 API design

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-1067) by [Unito](https://www.unito.io)

lukaszsliwa:
Resolves #283

- [x] Added tests RSL1a and RSL1b for both REST and Realtime clients
- [x] Ensure RSL1l, RSL1l1
- [x] Ensure RSL1h
- [x] Ensure RSL1k1, RSL1k2
- [x] Refactored `publish()` method in REST client
- [x] Refactored `publish()` method in Realtime client
- [x] Added `build_messages()` in `conversations.rb` and converting `name`, `data`, `attributes` to `Ably::Models::Message`

```ruby
# Publish a single message with (name, data) form (1)
channel.publish 'click', { x: 1, y: 2 }
      
# Publish a single message with single Hash form (2)
message = { name: 'click', data: { x: 1, y: 2 } }
channel.publish message
      
# Publish an array of message Hashes form (3)
messages = [
  { name: 'click', data: { x: 1, y: 2 } },
  { name: 'click', data: { x: 2, y: 3 } }
]
channel.publish messages
      
# Publish an array of Ably::Models::Message objects form (4)
messages = [
  Ably::Models::Message(name: 'click', data: { x: 1, y: 2 })
  Ably::Models::Message(name: 'click', data: { x: 2, y: 3 })
]
channel.publish messages
      
# Publish a single Ably::Models::Message object form (5)
message = Ably::Models::Message(name: 'click', data: { x: 1, y: 2 })
channel.publish message
      
```

### DIFF
--- a/lib/ably/modules/conversions.rb
+++ b/lib/ably/modules/conversions.rb
@@ -121,10 +121,11 @@ module Ably::Modules
     # @return [Array<Ably::Models::Message>]
     #
     def build_messages(name, data = nil, attributes = {})
-      if name.kind_of?(Enumerable)
-        return name.map { |item| Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes)) }
+      return [Ably::Models::Message(ensure_supported_name_and_payload(nil, data, attributes))] if name.nil?
+
+      Array(name).map do |item|
+        Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes))
       end
-      [Ably::Models::Message(ensure_supported_name_and_payload(name, data, attributes))]
     end
 
     # Ensures if the first argument (name) is a String, Hash or Ably::Models::Message object,
--- a/lib/ably/modules/conversions.rb
+++ b/lib/ably/modules/conversions.rb
@@ -115,5 +115,38 @@ module Ably::Modules
 
       raise Ably::Exceptions::UnsupportedDataType.new('Invalid data payload', 400, Ably::Exceptions::Codes::INVALID_MESSAGE_DATA_OR_ENCODING)
     end
+
+    # It converts the name, data, attributes into the array of Message objects
+    #
+    # @return [Array<Ably::Models::Message>]
+    #
+    def build_messages(name, data = nil, attributes = {})
+      if name.kind_of?(Enumerable)
+        return name.map { |item| Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes).dup) }
+      end
+      [Ably::Models::Message(ensure_supported_name_and_payload(name, data, attributes).dup)]
+    end
+
+    # Ensures if the first argument (name) is a String, Hash or Ably::Models::Message object,
+    # second argument (data) should be a String, Hash, Array or nil (see ensure_supported_payload() method).
+    #
+    # @return [Hash]        It contains :name, :data and other attributes
+    #
+    # (RSL1a, RSL1b)
+    #
+    def ensure_supported_name_and_payload(name, _data = nil, attributes = {})
+      return name.attributes if name.kind_of?(Ably::Models::Message)
+
+      data = _data
+      if (hash = name).kind_of?(Hash)
+        name, data = hash[:name], (hash[:data] || _data)
+        attributes.merge!(hash)
+      end
+
+      name = ensure_utf_8(:name, name, allow_nil: true)
+      ensure_supported_payload data
+
+      attributes.merge({ name: name, data: data })
+    end
   end
 end
--- a/lib/ably/modules/conversions.rb
+++ b/lib/ably/modules/conversions.rb
@@ -116,37 +116,37 @@ module Ably::Modules
       raise Ably::Exceptions::UnsupportedDataType.new('Invalid data payload', 400, Ably::Exceptions::Codes::INVALID_MESSAGE_DATA_OR_ENCODING)
     end
 
-    # It converts the name, data, attributes into the array of Message objects
+    # Converts the name, data, attributes into the array of Message objects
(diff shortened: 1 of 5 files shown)


lib/ably/modules/conversions.rb
```diff
@@ -115,5 +115,38 @@ def ensure_supported_payload(payload)
 
       raise Ably::Exceptions::UnsupportedDataType.new('Invalid data payload', 400, Ably::Exceptions::Codes::INVALID_MESSAGE_DATA_OR_ENCODING)
     end
+
+    # It converts the name, data, attributes into the array of Message objects
```
owenpearson, line 119: You should change this to "Converts the name, data, attributes into the array of Message objects"

lib/ably/modules/conversions.rb
```diff
@@ -115,5 +115,38 @@ def ensure_supported_payload(payload)
...
+      if name.kind_of?(Enumerable)
+        return name.map { |item| Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes).dup) }
+      end
+      [Ably::Models::Message(ensure_supported_name_and_payload(name, data, attributes).dup)]
+    end
+
+    # Ensures if the first argument (name) is a String, Hash or Ably::Models::Message object,
+    # second argument (data) should be a String, Hash, Array or nil (see ensure_supported_payload() method).
+    #
+    # @return [Hash]        It contains :name, :data and other attributes
```
owenpearson, line 133: Same here; remove the "It"

owenpearson (changes requested):
The code changes look good to me. Just needs a couple of small changes to inline documentation.

lib/ably/modules/conversions.rb
```diff
@@ -115,5 +115,38 @@ def ensure_supported_payload(payload)
 
       raise Ably::Exceptions::UnsupportedDataType.new('Invalid data payload', 400, Ably::Exceptions::Codes::INVALID_MESSAGE_DATA_OR_ENCODING)
     end
+
+    # It converts the name, data, attributes into the array of Message objects
+    #
+    # @return [Array<Ably::Models::Message>]
+    #
+    def build_messages(name, data = nil, attributes = {})
+      if name.kind_of?(Enumerable)
```
TheSmartnik, line 124: Since the only problem here is to support both array and single objects, I think it may be worthwhile to consider the following construction 

 ```ruby
def build_messages(name, data = nil, attributes = {})
  Array(name)e.map do |item|
    Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes).dup)
  end
end
```

It achieves the same thing and reads better. wdyt?

File: lib/ably/modules/conversions.rb [hunk 4]
```diff
@@ -115,5 +115,38 @@ def ensure_supported_payload(payload)
...
+        return name.map { |item| Ably::Models::Message(ensure_supported_name_and_payload(item, data, attributes).dup) }
```
TheSmartnik, line 125: Why do we need to dup message?

lib/ably/modules/conversions.rb
```diff
@@ -115,5 +115,38 @@ def ensure_supported_payload(payload)
...
+    end
+
+    # Ensures if the first argument (name) is a String, Hash or Ably::Models::Message object,
+    # second argument (data) should be a String, Hash, Array or nil (see ensure_supported_payload() method).
+    #
+    # @return [Hash]        It contains :name, :data and other attributes
+    #
+    # (RSL1a, RSL1b)
+    #
+    def ensure_supported_name_and_payload(name, _data = nil, attributes = {})
```
TheSmartnik, line 137: variable with an underscore by convention is used for unused variables. It's quite confusing to see `_data` being used later on

lib/ably/modules/conversions.rb
lukaszsliwa: Fixed

lib/ably/modules/conversions.rb
lukaszsliwa: Fixed

lib/ably/modules/conversions.rb
lukaszsliwa: Fixed

lib/ably/modules/conversions.rb
lukaszsliwa: Fixed

lib/ably/modules/conversions.rb
lukaszsliwa: Looks better but the above construction does not work.

For

```ruby
build_messages(nil, 'example data')
```

it should return

```ruby
[{"data": "example data"}]

```

but it returns empty array `[]`


`8` tests fail (most of them related to the `nil` name):

```spec

3101 examples, 8 failures, 10 pendings

Took 2052 seconds (34:12)
Tests Failed
\e[31m⚠ Note: Unit tests have also failed, but are not listed in the test failures above. Scroll up to the unit tests ⚠\e[0m
Unit tests exit code: 1
Acceptance tests exit code: 1

```

lukaszsliwa: Fixed b4acff6992ccc69f1ee74ca5b336f7986d7e95da

owenpearson (approved):
LGTM, thanks 👍

PR #342, opened 2022-01-31
Allowing ConnectionDetails#max_message_size

lukaszsliwa:
Allowing `max_message_size` for Realtime client. Client gets `max_message_size` from current `connection.details`. `ConnectionDetails` attributes are populated when connection is in `CONNECTED` state.

Added `max_message_size` to `Ably::Rest::Client`.

### DIFF
--- a/lib/ably/models/connection_details.rb
+++ b/lib/ably/models/connection_details.rb
@@ -21,6 +21,12 @@ module Ably::Models
   class ConnectionDetails
     include Ably::Modules::ModelCommon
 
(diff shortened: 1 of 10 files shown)


owenpearson (approved):
LGTM, thanks, although I would prefer to have a spec which mocks the server to respond with a different `maxMessageSize` and assert that the library sets its own `max_message_size`  accordingly. Do you reckon this would be easy enough to do @lukaszsliwa ?

PR #352, opened 2022-04-11
Rewrite logger unit spec to reduce flakiness

TheSmartnik:
* Adds `prevent_log_stubbing` to avoid `allow_any_instance_of` mock on Logger interfering with mocks from specs
* Rewrites spec to be shorter and more readable

I've rerun the test suite a few times and error don't appear anymore


owenpearson (approved):
Looks like a nice improvement, thanks 🙂

Issue #233, opened 2021-04-15
Error not emitted when failing to connect to an endpoint indefinitely


QuintinWillison:
This issue came out of an issue a customer had where there were no fallbacks for their configuration but there was no useful information output from the client library to logs to tell them that there had been a problem with their only available endpoint.

See [this Slack comment](https://ably-real-time.slack.com/archives/C01A3NPG99B/p1618482470017800) and its surrounding context.

PR #356, opened 2022-04-23
Add spec that on failed connection error is emitted

TheSmartnik:
Resolves: https://github.com/ably/ably-ruby/issues/233

The issue apparently has been already tackled but I've added extra spec to validate this


owenpearson (approved):
LGTM 👍

PR #361, opened 2022-05-10
Rspec junit format and observability server integration

lukaszsliwa:
- [x] Add `rspec_junit_formatter`
- [x] Fix run
- [x] Update  github workflow
- [x] Add observability integration


owenpearson (changes requested):
Looks good so far but:
- The CI doesn't seem to actually be running any of the tests right now, apparently they're all being filtered out?
- The shell script is quite hard to read, might be good to add some comments in there? I'd also be happy for you to bring in a command line argument parser as a dev dependency and write the script in ruby if that would be easier?

lukaszsliwa:
> Looks good so far but the shell script is quite hard to read, might be good to add some comments in there? I'd also be happy for you to bring in a command line argument parser as a dev dependency and write the script in ruby if that would be easier?

@owenpearson, I have removed the script and moved the command to the Github action run: https://github.com/ably/ably-ruby/blob/71da49baaa7063002672bc8d44ce20197ded377b/.github/workflows/check.yml#L29

The script was helpful befor but now we run 12 concurrent runs per push.

owenpearson (approved):
LGTM, thanks

PR #375, opened 2022-08-16
Fix doc comment: incorrect channel.state type

lukaszsliwa:
Fixed incorrect channel state class from https://github.com/ably/ably-ruby/blob/main/lib/ably/realtime/channel.rb#L29


owenpearson (approved):
LGTM

PR #376, opened 2022-09-02
[EDX-282] Changed related to docstring, generate docs and CI workflow

lukaszsliwa:
- [x] Add docs workflow (generate and upload documentation)
- [x] Update descriptions (base on the https://github.com/ably/canonical-api-reference-prototyping/blob/main/descriptions.md) 

https://sdk.ably.com/builds/ably/ably-ruby/pull/376/docs/


jamienewcomb:
@owenpearson we just need a review on the ruby syntax here please :)

lib/ably/rest/client.rb
```diff
@@ -357,20 +362,19 @@ def delete(path, params, options = {})
...
+      # handle features such as authentication, paging, fallback hosts, MsgPack and JSON support.
+      #
+      # @spec RSC19
       #
-      # @param method  [Symbol]    The HTTP method symbol such as +:get+, +:post+, +:put+
-      # @param path    [String]    The path of the URL such +/channel/foo/publish+
-      # @param params  [Hash, nil] Optional querystring params
-      # @param body    [Hash, nil] Optional body for the POST or PUT request, must be nil or a JSON-like object
-      # @param headers [Hash, nil] Optional additional headers
+      # @param method  [Symbol]    The request method to use, such as GET, POST.
```
owenpearson, line 371: I think we should keep the examples here as the actual symbols - `:get`, `:post`, etc

lib/ably/rest/client.rb
```diff
@@ -167,10 +166,10 @@ class Client
       # @return [Ably::Rest::Client]
       #
       # @example
-      #    # create a new client authenticating with basic auth
+      #    # The Ably API key or token string used to validate the client.
```
owenpearson, line 169: This seems wrong to me, the comment before was fine and the new one doesn't fit.

lib/ably/rest/client.rb
```diff
@@ -167,10 +166,10 @@ class Client
...
       #    client = Ably::Rest::Client.new('key.id:secret')
       #
-      #    # create a new client and configure a client ID used for presence
```
owenpearson, line 173: Also seems wrong, the example isn't using a `Hash` object

lib/ably/realtime/presence.rb
```diff
@@ -265,8 +277,9 @@ def subscribe(*actions, &callback)
       super
     end
 
-    # Unsubscribe the matching block for presence events on the associated Channel.
-    # If a block is not provided, all subscriptions will be unsubscribed
```
owenpearson, line 269: I think this line is useful and we should keep it

lib/ably/models/token_request.rb
```diff
@@ -14,10 +15,7 @@ def self.TokenRequest(attributes)
   end
 
 
-  # TokenRequest is a class that stores the attributes of a token request
-  #
-  # Ruby {http://ruby-doc.org/core/Time.html Time} objects are supported in place of Ably ms since epoch time fields.  However, if a numeric is provided
```
owenpearson, line 19: > Ruby Time objects are supported in place of Ably ms since epoch time fields

I think this means we will need to change a lot of these comments :/

lib/ably/models/paginated_result.rb
```diff
@@ -41,48 +44,57 @@ def initialize(http_response, base_url, client, options = {}, &each_block)
...
-    # @return [PaginatedResult,Ably::Util::SafeDeferrable]
     def first(&success_callback)
       async_wrap_if_realtime(success_callback) do
         return nil unless supports_pagination?
         PaginatedResult.new(client.get(pagination_url('first')), base_url, client, pagination_options, &each_block)
       end
     end
 
-    # Retrieve the next page of results.
-    # When used as part of the {Ably::Realtime} library, it will return a {Ably::Util::SafeDeferrable} object,
```
owenpearson, line 57: I think we shouldn't remove the bit about `SafeDeferrable`

lib/ably/models/idiomatic_ruby_wrapper.rb
```diff
@@ -35,9 +36,6 @@ module Ably::Models
   #   ruby_hash[:none] # => nil
   #   ruby_hash.none # => nil
   #
-  # @!attribute [r] stop_at
-  #   @return [Array<Symbol,String>] array of keys that this wrapper should stop wrapping at to preserve the underlying Hash as is
```
owenpearson, line 39: Is there a reason we're removing these? I think if we're not replacing them they should probably be kept

lib/ably/models/device_push_details.rb
```diff
@@ -20,38 +21,39 @@ def DevicePushDetails(device_push_details)
...
+    #
+    # @return [Symbol]
+    #
+    def state
+      attributes[:state]
+    end
 
-      define_method "#{attribute}=" do |val|
-        unless val.nil? || val.kind_of?(String)
-          raise ArgumentError, "#{attribute} must be nil or a string value"
```
owenpearson, line 49: What's the reason for this code change?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,36 @@
...
+          aws-region: eu-west-2
+          role-to-assume: arn:aws:iam::${{ secrets.ABLY_AWS_ACCOUNT_ID_SDK }}:role/ably-sdk-builds-ably-ruby
+          role-session-name: "${{ github.run_id }}-${{ github.run_number }}"
+
+      - name: Upload Documentation
+        uses: ably/sdk-upload-action@v1
+        with:
+          sourcePath: doc/
+          githubToken: ${{ secrets.GITHUB_TOKEN }}
+          artifactName: docs
```
owenpearson, line 36: missing newline at the end of this file

README.md
```diff
@@ -1,11 +1,15 @@
-# [Ably](https://ably.com)
```
owenpearson, line 1: We shouldn't be changing the README as part of this work. I'm guessing this content is supposed to appear in rubydoc API reference landing page, is there a way to change the content of that without changing the README?

owenpearson (changes requested):

.github/workflows/docs.yml
lukaszsliwa: Fixed

README.md
lukaszsliwa: Fixed: created INTRO.md and reverted README.md.

lib/ably/models/idiomatic_ruby_wrapper.rb
lukaszsliwa: IMO, `@!attribute [r] stop_at` and related are not so important. Because it's obvious for someone who read the code that there is stop_at attribute, so I focused on removed such yard comments. Also, it's not generated in the docs.

Okay, reverted.

lib/ably/models/paginated_result.rb
lukaszsliwa: Reverted.

lib/ably/rest/client.rb
lukaszsliwa: It's using Hash or String. Both ways are correct. I have fixed.

lib/ably/rest/client.rb
lukaszsliwa: Fixed

lib/ably/realtime/presence.rb
lukaszsliwa: Okay, fixed.

lib/ably/models/token_request.rb
lukaszsliwa: I have reverted back all similar comments.

owenpearson (approved):
LGTM

PR #381, opened 2022-10-03
test: remove deprecated recover behaviour spec

owenpearson:
See [internal slack thread](https://ably-real-time.slack.com/archives/C02P7F81DPS/p1664792520213559)

### DIFF
--- a/spec/acceptance/realtime/connection_spec.rb
+++ b/spec/acceptance/realtime/connection_spec.rb
@@ -1620,21 +1620,6 @@ describe Ably::Realtime::Connection, :event_machine do
           end
         end
 
-        context 'with invalid formatted value sent to server' do
-          let(:client_options) { default_options.merge(recover: 'not-a-valid-connection-key:1:0', log_level: :none) }
-
-          it 'sets the #error_reason and moves the connection to FAILED' do
-            connection.once(:failed) do |state_change|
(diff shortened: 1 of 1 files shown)


QuintinWillison (approved):
To explain the context of this change for those looking at this pull request without access to our internal Slack... @owenpearson had asked this question of those working in our 'Capacity' pod:

> [this test](https://github.com/ably/ably-ruby/blob/9cb6d6b560bd1533af3e194602f6f857eca99968/spec/acceptance/realtime/connection_spec.rb#L1623-L1636) has recently started failing consistently in ably-ruby, i’m guessing it’s related to the work yall are doing here. could you verify if that’s correct and, if so, do you have a recommendation of how we should fix it or if i should just remove it entirely?

to which @SimonWoolf replied with:

> Ah, yup, that is related.
> So: we used to distinguish between malformed connection key (realtime responds with an ERROR) and unknown connection key (realtime responds with CONNECTED but with a new connection and an error member, as long as the auth still checks out fine etc). But with multiple styles of connectionKey now, we now just have a single failure case, if the connection key isn't good then we always give you a CONNECTED with an error member
> I'd just remove that test entirely, it was testing server behaviour more than sdk behaviour. as long as there's still a test somewhere that 'an ERROR moves the connection to a failure state' which there presumably still is

PR #382, opened 2022-10-03
fix: remove inbound message size validation

### LINKED ISSUE #377: Incorrect ProtocolMessage#connection_details object (overwrites original connection_details send on CONNECTED state)
See [internal slack thread](https://ably-real-time.slack.com/archives/C01A3NPG99B/p1662719301004219)

A customer has reported that the library emits `Channel message limit exceeded xxx bytes` errors when the number of bytes in the message is less than the MAX_MESSAGE_SIZE on the account. It appears that the library is using the default MAX_MESSAGE_SIZE in favour of the maxMessageSize received from ConnectionDetails.

## Update 2022-09-13
ProtocolMessage.connection_details object returns incorrect connection details object. [https://github.com/ably/ably-ruby/blob/7cd9226bfe43532a44614fe6b8fc063400e372c1/lib/ably/models/protocol_message.rb#L257](https://github.com/ably/ably-ruby/blob/7cd9226bfe43532a44614fe6b8fc063400e372c1/lib/ably/models/protocol_message.rb#L257|smart-link) 

It should delegate/reference to the connection_details sent on CONNECTED state.

attributes[:connection_details] is nil when sending new messages (after CONNECTED) and it overwrites the connection details object.

## Update 2022-09-21
Connection details are missed between messages in websocket transport: [https://github.com/ably/ably-ruby/blob/main/lib/ably/realtime/connection/websocket_transport.rb#L209,L227](https://github.com/ably/ably-ruby/blob/main/lib/ably/realtime/connection/websocket_transport.rb#L209,L227|smart-link). When event_data  contains connection_details it should be saved/updated and filled to the protocol message and shared between requests.

ProtocolMessage#connection_details should be replaced with attr_reader :connection_details and assigned in initializer.

┆Issue is synchronized with this [Jira Uncategorised](https://ably.atlassian.net/browse/SDK-2629) by [Unito](https://www.unito.io)
┆Attachments: <a href="https://ably.atlassian.net/rest/api/2/attachment/content/12781">protocol_message_connection_details_debug.log</a>

owenpearson:
Resolves #377

`TO3l8` specifies the requirement of the SDK to validate message size for outbound messages so the check for inbound messages is not needed. Moreover the existing implementation reads `connection_details` from the `MESSAGE` protocol message (which is always unset) and falls back to the library default, so doesn't work when an account has a configured max_message_size

### DIFF
--- a/lib/ably/models/protocol_message.rb
+++ b/lib/ably/models/protocol_message.rb
@@ -189,10 +189,6 @@ module Ably::Models
       presence.map(&:size).sum + messages.map(&:size).sum
     end
 
-    def has_correct_message_size?
-      message_size <= connection_details.max_message_size
-    end
-
     def params
       @params ||= attributes[:params].to_h
     end
--- a/lib/ably/realtime/client/incoming_message_dispatcher.rb
+++ b/lib/ably/realtime/client/incoming_message_dispatcher.rb
@@ -121,23 +121,15 @@ module Ably::Realtime
             presence.manager.sync_process_messages protocol_message.channel_serial, protocol_message.presence
 
           when ACTION.Presence
-            if protocol_message.has_correct_message_size?
-              presence = get_channel(protocol_message.channel).presence
-              protocol_message.presence.each do |presence_message|
-                presence.__incoming_msgbus__.publish :presence, presence_message
-              end
-            else
-              logger.fatal Ably::Exceptions::ProtocolError.new("Not published. Channel message limit exceeded #{protocol_message.message_size} bytes", 400, Ably::Exceptions::Codes::UNABLE_TO_RECOVER_CHANNEL_MESSAGE_LIMIT_EXCEEDED).message
+            presence = get_channel(protocol_message.channel).presence
+            protocol_message.presence.each do |presence_message|
+              presence.__incoming_msgbus__.publish :presence, presence_message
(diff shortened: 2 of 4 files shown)


lukaszsliwa (approved):

QuintinWillison (changes requested):
I see we're mostly removing code here.

What I can't see is _why_ we're making this change. There is no opening comment on this pull request that explains to reviewers or those looking back in future why the change is being made. Even the [single commit](https://github.com/ably/ably-ruby/pull/382/commits/e2764e6c08c04f733109abdcdc929b3602555218) only contains a single line:

> fix: remove inbound max_message_size check

which pretty much matches the pull request title:

> fix: remove inbound message size validation

What I would like to see, one or more of:

- Opening comment explaining why the change is being made (i.e. what is wrong in the current implementation)
- A `bug` label assigned to ensure it makes it into the changelog as such
- A link to a GitHub issue that is `bug` labelled, describing the problem with the current implementation
- A link from this PR, or the linked GitHub issue, to other sources of information ... e.g. an " (internal)"-suffixed link to a Slack thread, perhaps

owenpearson:
Good point @QuintinWillison, I've updated the description now

QuintinWillison (approved):

PR #397, opened 2024-04-24
[ECO-4767] Fix server certificate validation for HTTP in Realtime

### LINKED ISSUE #396: Invalid server certificate error for `internet-up` endpoint
@sacOO7 noticed that the acceptance tests are failing (e.g. [here](https://github.com/ably/ably-ruby/actions/runs/8799287951/job/24148038958?pr=391)) with the following error:

```
     OpenSSL::SSL::SSLError:
       unable to verify the server certificate for "internet-up.ably-realtime.com"
```

Not sure how long this has been happening for.

@lmars mentioned this might be to do with SNI; when the client doesn’t provide SNI then the server responds with a certificate that doesn’t include `internet-up.ably-realtime.com` as an allowed subject name. It’s quite possible that we are not supplying SNI; need to look into that. As for why this used to work, perhaps we changed the way that we generate certs so that they now only work when SNI is supplied, not sure (flagged this to infrastructure support).

[Internal Slack thread](https://ably-real-time.slack.com/archives/C03JDBVM5MY/p1713870728383829).

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4767) by [Unito](https://www.unito.io)

lawrence-forooghian:
Resolves #396.

(The tests pass when combined with @sacOO7’s WIP test fixes from #391; see #403.)

### DIFF
--- a/ably.gemspec
+++ b/ably.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
 
   spec.add_runtime_dependency 'eventmachine', '~> 1.2.6'
-  spec.add_runtime_dependency 'em-http-request', '~> 1.1'
+  spec.add_runtime_dependency 'ably-em-http-request', '~> 1.1.8'
   spec.add_runtime_dependency 'statesman', '~> 9.0'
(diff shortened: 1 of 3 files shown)


ttypic (approved):
LGTM

PR #398, opened 2024-04-25
Turn off `continue-on-error` in CI

lawrence-forooghian:
We want the workflow run to fail if any of the jobs fails.

The motivation for the current configuration is not clear to me, but I’m guessing the idea was that the failure of a single job shouldn’t cancel the others. Turning off fail-fast should do the same thing.

(Copied from ably-js commit ddef9f7.)

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -7,8 +7,8 @@ on:
 jobs:
   check:
     runs-on: ubuntu-latest
-    continue-on-error: true
     strategy:
+      fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


lawrence-forooghian:
I suggest that we merge this one even though CI is failing (given that the tests were always failing but it was just hidden before).

sacOO7:
@lawrence-forooghian is this implemented in other SDKs as well?

lawrence-forooghian:
> @lawrence-forooghian is this implemented in other SDKs as well?

As mentioned above, we do it in ably-js.

PR #405, opened 2024-05-20
refactored string check for either nil or empty string

sacOO7:


sacOO7:
Closing in favor of #410

PR #410, opened 2024-06-06
[ECO-4686][Protocol-2] Implement internal presence

### LINKED ISSUE #392: [ably-ruby][no-connection-serial] Implement missing internal presence
* Implement internal presence and presence queue
* Conform to presence spec starting from https://sdk.ably.com/builds/ably/specification/main/features/#RTP1
* Make sure implementation is thread safe

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4686) by [Unito](https://www.unito.io)

sacOO7:
- Fixed #392 
 - Improvement of previous closed PR https://github.com/ably/ably-ruby/pull/405
  
  **PR description** 
- renamed `:in_sync` enum to `:sync_completed`, since `:in_sync` is a very confusing term and doesn't go as per spec.
 - removed `resume_sync_proc` code that does client initiated sync when channel attaches with `has_presence` flag. This doesn't go as per spec. 
 - added `on_attach` method to `presence_manager` that 
 1. start or end sync depending on `has_presence` flag
 2. enter local members when `on_attach` method is called.
 - Added a new `enter_client_with_id` method to presence that enters local members with `id`. Earlier local members were entered at the end of the sync and didn't support entering members with given id.

### DIFF
--- a/lib/ably/realtime/presence/members_map.rb
+++ b/lib/ably/realtime/presence/members_map.rb
@@ -152,6 +152,29 @@ module Ably::Realtime
         @local_members
       end
 
+      def enter_local_members
+        local_members.values.each do |member|
+          local_client_id = member.client_id || client.auth.client_id
+          logger.debug { "#{self.class.name}: Manually re-entering local presence member, client ID: #{local_client_id} with data: #{member.data}" }
+          presence.enter_client_with_id(member.id, local_client_id, member.data).tap do |deferrable|
+            deferrable.errback do |error|
+              presence_message_client_id = member.client_id || client.auth.client_id
+              re_enter_error = Ably::Models::ErrorInfo.new(
+                message: "unable to automatically re-enter presence channel for client_id '#{presence_message_client_id}'. Source error code #{error.code} and message '#{error.message}'",
+                code: Ably::Exceptions::Codes::UNABLE_TO_AUTOMATICALLY_REENTER_PRESENCE_CHANNEL
+              )
+              channel.emit :update, Ably::Models::ChannelStateChange.new(
+                current: channel.state,
+                previous: channel.state,
+                event: Ably::Realtime::Channel::EVENT(:update),
+                reason: re_enter_error,
+                resumed: true
+              )
+            end
+          end
+        end
+      end
+
       private
       attr_reader :sync_session_id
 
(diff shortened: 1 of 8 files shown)


lib/ably/realtime/presence/members_map.rb
```diff
@@ -233,60 +224,31 @@ def setup_event_handlers
...
-            channel.emit :update, Ably::Models::ChannelStateChange.new(
-              current: channel.state,
-              previous: channel.state,
-              event: Ably::Realtime::Channel::EVENT(:update),
-              reason: re_enter_error,
-              resumed: true
-            )
+      def enter_local_members
+        local_members.values.each do |member|
+          local_client_id = member.client_id || client.auth.client_id
```
lawrence-forooghian, line 231: Why `|| client.auth.client_id` here? [RTP17g](https://sdk.ably.com/builds/ably/specification/main/features/#RTP17g) just says "using the `clientId` … attributes from that member".

lib/ably/realtime/presence/presence_manager.rb
```diff
@@ -19,13 +19,18 @@ def initialize(presence)
...
+        if has_presence_flag
+          # Expect SYNC ProtocolMessages from the server with a list of current members on this channel
+          presence.members.change_state :sync_starting
+        else
+          # There server has indicated that there are no SYNC ProtocolMessages to come because
+          # there are no members on this channel
+          logger.debug { "#{self.class.name}: Emitting leave events for all members as a SYNC is not expected and thus there are no members on the channel" }
+          presence.members.change_state :sync_none
+        end
+        presence.members.send(:enter_local_members)
```
lawrence-forooghian, line 33: Why not `presence.members.enter_local_members`?

lawrence-forooghian:
And, same question as in https://github.com/ably/ably-ruby/pull/409#pullrequestreview-2125506559 re tests?

lib/ably/realtime/presence/presence_manager.rb
sacOO7: Actually, `enter_local_members` is a private method. Since, we don't want to expose it as a public method, we have marked it as private. If try to call private method directly, ruby throws runtime error.

sacOO7:
> And, same question as in [#409 (review)](https://github.com/ably/ably-ruby/pull/409#pullrequestreview-2125506559) re tests?

Tests fixed as a part of https://github.com/ably/ably-ruby/pull/411

lib/ably/realtime/presence/presence_manager.rb
lawrence-forooghian: We really shouldn’t be having one class calling a private method of another class; that doesn't seem right. I mean, in this case `enter_local_members` is a private method on `MembersMap`, which is a class that isn't even part of the public API of the library, right? So I think it'd be fine for `enter_local_members` to be public?

sacOO7: Fixed https://github.com/ably/ably-ruby/pull/410/commits/446a16fabb036ff135b765f0bcce5a88bf09b11e

lib/ably/realtime/presence/members_map.rb
sacOO7: Fixed https://github.com/ably/ably-ruby/pull/418/commits/c10e485344ac4f87e7d83445a6ab3b9b287f1129

lawrence-forooghian (approved):

Issue #413, opened 2024-06-20
Improper use of EventEmitter [violation of RTC10]


sacOO7:
- As per [RTC10](https://sdk.ably.com/builds/ably/specification/main/features/#RTC10), library must not register any listeners for internal use. 
- At most of the places, ably-ruby uses emitter callbacks to detect state changes and execute internal code.
- This is causing problems in order of code execution with respect to internal and external code. Same is being reflected in flaky behavior of the tests.
- We have fixed one of the similar issue as a part of https://github.com/ably/ably-ruby/pull/411/commits/0e548c27b6b8560c477b3aeda9fd6af72f040074.
- Aim of this issue is to find similar internal emitter callback usages and fix them.

sacOO7:
One of the solution would be to use internalEmitter that's specific to internal library. 
We use the same in case of `ably-go`
https://github.com/ably/ably-go/blob/0885021e9c92c9895dfbeb644e71fa45936cd4ae/ably/realtime_conn.go#L1016 and
https://github.com/ably/ably-go/blob/0885021e9c92c9895dfbeb644e71fa45936cd4ae/ably/realtime_channel.go#L1028

PR #416, opened 2024-06-24
[ECO-4687][Protocol-2] Fix duplicate outgoing ATTACH msg

sacOO7:
Main issue was related to the fact that, every msg that needs to be sent is queued first in the global queue and then sent one by one. Since, protocol 2 supports sending force ATTACH on every connected msg, if there is existing ATTACH in queue, both of the ATTACH will be sent on CONNECTED -> https://github.com/ably/specification/pull/193/files

Fixed ably-ruby implementation responsible for sending duplicate ATTACH message.
1. Send attach only when connection state is CONNECTED
2. Do not queue the attach message, instead send immediately on connected
3. Removed use of emitter pattern while clearing attach/detach timers (see https://github.com/ably/ably-ruby/issues/413)

Link to internal discussion 
https://ably-real-time.slack.com/archives/C8SPU4589/p1718275991214299?thread_ts=1718115075.720879&cid=C8SPU4589 ( check for last few messages regarding decision taken ) 

PS. It will be easy to understand changes by going through each commit.


lib/ably/realtime/channel/channel_manager.rb
```diff
@@ -226,11 +232,6 @@ def send_state_change_protocol_message(new_state, state_if_failed, message_optio
...
         end
 
-        channel.once_state_changed do
-          @pending_state_change_timer.cancel if @pending_state_change_timer
-          @pending_state_change_timer = nil
-        end
-
         resend_if_disconnected_and_connected = lambda do
           connection.unsafe_once(:disconnected) do
             next unless pending_state_change_timer
```
sacOO7, line 237: @lawrence-forooghian I didn't touch the existing code as such.
But I am having hard time understanding usage of `next unless` here.
Since, next is only used inside a loop, I think it is being used to inform internal emitter callback execution loop to execute next callback when condition holds true. Basically, next lines of code is skipped when `pending_state_change_timer` is nil. I think it can be simplified using simple `if` condition.

sacOO7: I have kept changes small for better understanding. Maybe, I can refactor the code as a part of next commit.

sacOO7: https://github.com/ably/ably-ruby/pull/416/commits/1324a72a858910813514ae65bd66dd29b13adff7

sacOO7:
I have kept implementation for sending `ATTACH` and `DETACH` msg separate as a part of https://github.com/ably/ably-ruby/pull/416/commits/8b9b556e878b7871611ee6f621f72e9cbd229707. 
Earlier, there most of the code was `DETACH` specific and had nothing to do with sending `ATTACH` message.

lib/ably/realtime/connection.rb
```diff
@@ -439,9 +439,22 @@ def send_protocol_message(protocol_message)
         end
       end
 
+      def send_protocol_message_immediately(protocol_message)
+        Ably::Models::ProtocolMessage.new(protocol_message, logger: logger).tap do |message|
```
lawrence-forooghian, line 443: Why `#tap`? That's usually used when you want to create an object and also return it. Can you just do

```ruby
message = Ably::Models::ProtocolMessage.new(protocol_message, logger: logger)
add_message_to_outgoing_queue(message, true)
(and so on)
```

lawrence-forooghian:
I've tried taking a look at this PR but it's quite complicated and I don't really understand what's going on. Might be best to have a call about it.

lawrence-forooghian:
Recapping our call:

- let's move the connection state check to `#attach` per RTL4i, then we won't need this new `force_attach` flag
- the final commit is just a refactor; splits the attach and detach logic (which deviated further in the first commit) into separate methods instead of a shared method
- I asked whether we need some new tests? Sachin said he'd look into it
- I still don't understand the `notify_state_change` method, or what it has to do with #413, or what #413 has to do with this issue; we need to have another conversation about that

lib/ably/realtime/connection.rb
sacOO7: Fixed https://github.com/ably/ably-ruby/pull/416/commits/7a973ae7718d2019544a61ddfe11e115cc0654ac

sacOO7:
> Recapping our call:
> 
> * let's move the connection state check to `#attach` per RTL4i, then we won't need this new `force_attach` flag
> * the final commit is just a refactor; splits the attach and detach logic (which deviated further in the first commit) into separate methods instead of a shared method
> * I asked whether we need some new tests? Sachin said he'd look into it
> * I still don't understand the `notify_state_change` method, or what it has to do with [Improper use of EventEmitter [violation of RTC10] #413](https://github.com/ably/ably-ruby/issues/413), or what [Improper use of EventEmitter [violation of RTC10] #413](https://github.com/ably/ably-ruby/issues/413) has to do with this issue; we need to have another conversation about that

- I have removed use of `forced_attach` from the code and moved `connected` check while calling `attach` explicitly.
- About `notify_state_change` => this is introduced to replce -> https://github.com/ably/ably-ruby/blob/main/lib/ably/realtime/channel/channel_manager.rb#L228-L231.
- As, you can see 
```
        channel.once_state_changed do
          @pending_state_change_timer.cancel if @pending_state_change_timer
          @pending_state_change_timer = nil
        end
```
is registering a callback on `emitter` class, same is used to register callbacks externally
e.g.
```
    channel.attach do
       # does some stuff for longer time
    end
```
- All of these callbacks gets executed in the order they are registered. You can see the callback array ->
https://github.com/ably/ably-ruby/blob/7f156b7e92aa26126d2d253a962a40e965988c09/lib/ably/modules/event_emitter.rb#L162 and it gets executed in a for loop -> https://github.com/ably/ably-ruby/blob/7f156b7e92aa26126d2d253a962a40e965988c09/lib/ably/modules/event_emitter.rb#L91
- So, if we register an external callback first that does work for long time, the callback responsible for cancelling timer won't execute. Hence, it will take channel state into suspended state.
- This is a one of the problem described as per [RTC10](https://sdk.ably.com/builds/ably/specification/main/features/#RTC10) ( do read related spec )
- So, this is a small fix done to avoid sending duplicate ATTACH message, since going into `suspended` state again triggers reattach mechanism as per [RTL4f](https://sdk.ably.com/builds/ably/specification/main/features/#RTL4f)

lawrence-forooghian (approved):

PR #420, opened 2024-07-01
Push docstrings updates

owenpearson:
updates GCM to FCM (GCM was deprecated in favour of FCM), and adds web as a mentioned push target.


sacOO7 (approved):
LGTM

PR #421, opened 2024-07-02
[ECO-4847] Fix until attach serial test

### LINKED ISSUE #415: Failing Test - updates attach_serial
- [updates attach_serial test](https://github.com/ably/ably-ruby/blob/062d98f98d4c67a7423ad7a4bb5856b3621e4f71/spec/acceptance/realtime/channel_history_spec.rb#L197) test started failing because of server side changes made to the channel serial.
- [This test is skipped on the main branch](https://github.com/ably/ably-ruby/blob/062d98f98d4c67a7423ad7a4bb5856b3621e4f71/spec/acceptance/realtime/channel_history_spec.rb#L197) because it was confirmed to be a bad test in the first place.
- Link to the conversation -> https://ably.atlassian.net/browse/REA-1910

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4847) by [Unito](https://www.unito.io)

sacOO7:
Fixed #415

### DIFF
--- a/spec/acceptance/realtime/channel_history_spec.rb
+++ b/spec/acceptance/realtime/channel_history_spec.rb
@@ -184,26 +184,22 @@ describe Ably::Realtime::Channel, '#history', :event_machine do
       end
 
       context 'when channel receives update event after an attachment' do
-        attach_serial = "old attach serial"
+        old_attach_serial = ""
         new_attach_serial = "xxxx-xxxx-1"
         before do
           channel.on(:attached) do
-            attach_serial = channel.properties.attach_serial
-            channel.publish(event, message_after_attach) do
-              attached_message = Ably::Models::ProtocolMessage.new(action: 11, channel: channel_name, flags: 0, channel_serial: new_attach_serial)
-              client.connection.__incoming_protocol_msgbus__.publish :protocol_message, attached_message
-            end
+            old_attach_serial = channel.properties.attach_serial
+            attached_message = Ably::Models::ProtocolMessage.new(action: 11, channel: channel_name, flags: 0, channel_serial: new_attach_serial)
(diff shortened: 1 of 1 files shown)


spec/acceptance/realtime/channel_history_spec.rb
```diff
@@ -184,8 +184,11 @@ def ensure_message_history_direction_and_paging_is_correct(direction)
       end
 
       context 'when channel receives update event after an attachment' do
```
lawrence-forooghian, line 186: What spec point is this test for? I don't know what "when channel recieves update event" means. The only spec point that describes how `attach_serial` should be updated is RTL15a, which just says that

>  attachSerial is unset when the channel is instantiated, and is updated with the channelSerial from each ATTACHED ProtocolMessage received from Ably with a matching channel attribute

So if this is a test for RTL15a, then I don't understand why the test does all this stuff like `rest_channel.publish event, message_before_attach`…

sacOO7: Generally, `channelSerial` has a format `xxxx-xxx-0`, then last digit is incremented for subsequent serials. Currently, we are setting it to `1` from `0`.
Maybe, I will update the test so it's easy to understand 👍

lawrence-forooghian: But the exact value of the `channelSerial` doesn't matter at all, does it?

sacOO7: Okay, updated the code https://github.com/ably/ably-ruby/pull/421/commits/ba82881d667e842a3f10901504723806fdbeadc6

lawrence-forooghian (approved):

Issue #425, opened 2024-07-05
Fix tests based on more clarification on RSA7e


sacOO7:
- [RSA7e](https://sdk.ably.com/builds/ably/specification/main/features/#RSA7e) specifies if `clientOptions.clientId and basic auth provided, set clientId as a header/queryparam` although server accepts `clientId` on connection for token auth.
- Link to internal conversation -> https://ably-real-time.slack.com/archives/C8SPU4589/p1720167508719079
- In favor of seeking more clarity, two tests are skipped
1. Ably::Realtime::Connection using JSON protocol intialization with token auth with opaque token string that contain an implicit client_id string that is incompatible with the current client client_id fails the connection
2. Client using MsgPack protocol initialization token auth with a Proc for the :auth_callback option with a wildcard client_id token and an explicit client_id in ClientOptions allows uses the explicit client_id in the connection

┆Issue is synchronized with this [Jira Task](https://ably.atlassian.net/browse/ECO-4865) by [Unito](https://www.unito.io)

sacOO7:
Previous related issue https://github.com/ably/ably-ruby/issues/414

sacOO7:
Clarification using spec PR -> https://github.com/ably/specification/pull/195

Issue #378, opened 2022-09-13
[ably-ruby][no-connection-serial] Implement code + tests for no-connection-serial


jamienewcomb:
Implement the changes listed in https://github.com/ably/docs/pull/1526
Reference impl -> https://github.com/ably/ably-js/pull/1045

┆Issue is synchronized with this [Jira Story](https://ably.atlassian.net/browse/ECO-4058) by [Unito](https://www.unito.io)

PR #426, opened 2024-07-05
[ECO-4058][Protocol-2] Feature/integration protocol 2 tests

sacOO7:
- Related to #378 
- Most of the changes are related to `Making sure connection state is connected before attaching channel` because if reduces flakiness of the tests
- Two skipped tests will be fixed as a part of https://github.com/ably/ably-ruby/issues/425 since they are not part of protocol 2 work.

### DIFF
--- a/spec/acceptance/realtime/client_spec.rb
+++ b/spec/acceptance/realtime/client_spec.rb
@@ -133,7 +133,7 @@ describe Ably::Realtime::Client, :event_machine do
             end
           end
 
-          context 'with a wildcard client_id token' do
+          context 'with a wildcard client_id token ' do
             subject                 { auto_close Ably::Realtime::Client.new(client_options) }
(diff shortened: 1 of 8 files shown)


spec/acceptance/realtime/client_spec.rb
```diff
@@ -142,7 +142,7 @@
             context 'and an explicit client_id in ClientOptions' do
               let(:client_id) { random_str }
 
-              it 'allows uses the explicit client_id in the connection' do
+              xit 'allows uses the explicit client_id in the connection' do
```
ttypic, line 145: Could you add comment for skipped tests that they will be fixed as a part of https://github.com/ably/ably-ruby/issues/425? (I think skipped test always should be with comment describing why they were skipped and when we are going to fix them)

sacOO7: Sure

ttypic (approved):
Other than comments for skipped test looks great! Thank you!

PR #427, opened 2024-07-05
Release 1.2.6

sacOO7:
- Added support for protocol 2 / no-connection serial.
- Refactored code and removed old implementation for client initiated presence sync.
- Added `createRecoveryKey` method to generate recoveryKey for new connection.


ttypic (approved):
Nice work!

PR #431, opened 2024-07-08
Fix lamda used in channel detach

sacOO7:
- Fix lamda block supplied as a part of channel detach, since use of yield inside lamda is not allowed.
- Instead block should be passed and called explicitly.
- https://stackoverflow.com/questions/30649742/declaring-a-ruby-lambda-with-a-yield


ttypic (approved):
LGTM

PR #432, opened 2024-07-23
spec: Don't assert error messages

lmars:
They are subject to change, so checking the error code is enough to know that the expected error occurred.


sacOO7 (approved):
LGTM

PR #435, opened 2024-08-21
spec: Don't expect integers to round-trip through msg.extras

lmars:
Internally, we store msg.extras in a protobuf.Struct which stores a JSON number in a double field:

https://github.com/protocolbuffers/protobuf/blob/5cbf13bc9ab0f883bcb664241f6be8d962915569/src/google/protobuf/struct.proto#L56-L78

This means that when a connection is using binary encoding (i.e. msgpack), any numbers in msg.extras are stored as floats even if the inbound message encoded the number as an int type, so the inbound and outbound representations are not guaranteed to be the same, although the semantic numeric value will be the same.

This commit handles this nuance by not expecting an integer to come back as an integer in msg.extras, but expecting a float to instead.

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated test data to ensure type consistency by changing an integer value to a floating-point number, enhancing accuracy in data processing scenarios.
  
- **Tests**
	- Improved test reliability by refining the data structure used in acceptance tests, which may impact validations and assertions involved in encoding and decoding operations.

### DIFF
--- a/spec/acceptance/realtime/message_spec.rb
+++ b/spec/acceptance/realtime/message_spec.rb
@@ -173,7 +173,7 @@ describe 'Ably::Realtime::Channel Message', :event_machine do
       end
 
       context 'JSON Array' do
-        let(:data) { { 'push' => { 'data' => { 'key' => [ true, false, 55, nil, 'string', { 'Hash' => true }, ['array'] ] } } } }
+        let(:data) { { 'push' => { 'data' => { 'key' => [ true, false, 55.1, nil, 'string', { 'Hash' => true }, ['array'] ] } } } }
 
         it 'is encoded and decoded to the same Array' do
           publish_and_check_extras data


owenpearson (commented):
I'm not entirely sure about this, probably not a big deal for frontdoor websockets but will this affect extras when publishing push notifications through frontdoors? for example we have live docs which specify an int value in extras [here](https://ably.com/docs/push/publish#apns-headers), is it safe to assume this will all still work correctly if we stop enforcing that ints survive protobuf encoding?

lmars:
@owenpearson so this only affects msgpack encoding, JSON doesn't have different number types so whether we store an integer as a float or int internally, in JSON it will encode the same (and specifically, this means it won't affect push notifications since they will be encoded as JSON when we send them to the push platform, similarly for integrations), so I don't think this will impact anything other than clients which subscribe using msgpack always getting numbers as floats in msgpack-encoded extras.

owenpearson (approved):
ok nice, sgtm

PR #443, opened 2025-06-10
Fix decoding of MessagePack data, and other fixes to CI

lawrence-forooghian:
This fixes:

- an issue with decoding MessagePack data due to a breaking change in the `websocket-driver` gem (we should do a release for this because it may be affecting customers)
- a partial fix for #444, to prevent token errors from triggering the use of a fallback host for realtime connections, which fixes an intermittent test failure
- a test that was failing because of backend changes
- a couple of other things that were causing issues in CI

See commit messages for more details.


ttypic (approved):
LGTM

sacOO7 (approved):
lgtm

PR #451, opened 2026-05-26
ci: GHA workflow security cleanup

emptyhammond:
Routine hygiene pass over the GitHub Actions workflows in this repo, addressing findings from a workflow security audit. Changes are split into five commits, one per finding type:
- Disable credential persistence on actions/checkout steps so the workflow's GITHUB_TOKEN is not left in the local git config after checkout.
- Scope each job's permissions explicitly: top-level permissions: {}, with each job granted only the GITHUB_TOKEN scopes it actually needs.
- Move the workflow-context expansion (${{ matrix.type }}) out of run: shell source and read it from the already-defined TEST_TYPE env binding instead.
- Pin all third-party actions to commit SHAs (with the tag preserved as a comment) so an upstream tag move can't silently change what runs in CI.
- Replace secrets: inherit on the reusable ably/features workflow call with an explicit passthrough of the single secret it declares (ABLY_AWS_ACCOUNT_ID_SDK), so unrelated repo secrets aren't forwarded into it.

No behavioural changes intended — the workflows run the same checks against the same inputs.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -19,11 +19,11 @@ jobs:
         protocol: [ 'json', 'msgpack' ]
         type: [ 'unit', 'acceptance' ]
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@ee0669bd1cc54295c223e0bb666b733df41de1c5 # v2
         with:
           submodules: 'recursive'
           persist-credentials: false
-      - uses: ruby/setup-ruby@v1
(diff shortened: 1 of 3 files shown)


lawrence-forooghian:
@emptyhammond looks like this has broken something in CI

emptyhammond:
> @emptyhammond looks like this has broken something in CI

@lawrence-forooghian the failures seem to be related to presence acceptance tests - not the workflow changes.

lawrence-forooghian (approved):

PR #453, opened 2026-09-01
Split Pub/Sub SDK into core and server gems

umair-ably:
Implements the PubSub package split (PDR-091b) for the Ruby SDK, mirroring ably/ably-js#2293.

* **New public package**: `ably-pubsub-server` (`server/`), allowing Pub/Sub connections to explicitly declare they are server-side for realtime classification and MAU billing. Ruby is a server-only SDK under PDR-091b2, so there is no device gem. Factory functions are the only recommended entry points:
  * `Ably::PubSub::Server.create_http_client(...)`
  * `Ably::PubSub::Server.create_realtime_client(...)`
* **Shared core**: the implementation moves to `ably-pubsub-core` (`core/`), for internal/shared usage and not intended as a public API. The ably-common submodule moves with it.
* **Agent stamping**: the factories append the versionless side-declaring agent entry `ably-pubsub-server` (a pure flag, matching its ably/ably-common#361 registration — the `ably-ruby/x.y.z` entry beside it carries the version) via a new additive `:agents` client option. Caller entries are preserved; the side entry wins a collision on its own identifier. The `-server` suffix is what MAU classification reads, so specs assert it loudly.
* **Lockstep versioning**: core and server are both at 2.0.0, with the server gem pinning core to the exact version.
* The old `ably` gemspec is removed: the `ably` gem enters its 1-year maintenance window from a maintenance branch, not from here.

Unit suite green (1250 examples), sandbox acceptance smoke-tested, both gems build.


ttypic (approved):
LGTM

PR #454, opened 2026-09-01
Add lockstep release workflow and migration docs

umair-ably:
Stacked on #453 (review that first — this diff shows only the release-tooling changes).

Implements the release-tooling part of the PubSub package split (PDR-091b):

* **`release.yml`**: one `workflow_dispatch` releases `ably-pubsub-core` and `ably-pubsub-server` at the same version via [RubyGems trusted publishing](https://guides.rubygems.org/trusted-publishing/) (OIDC — no long-lived API keys).
  * A **pre-flight** fails before anything is pushed if the version input, either gem's version constant, or the server gem's exact `= x.y.z` core pin disagree — a partial or skewed release is impossible.
  * **Reversible on failure**: if the server push fails after the core push succeeded, re-running with the same version skips the already-published gem.
  * The server gem is only pushed once the core version is visible on RubyGems, so the exact pin always resolves.
* **CONTRIBUTING.md**: documents the two-gem layout and replaces the manual `rake release` process (and the `ably-ruby-rest` release step — both legacy gems are maintenance-only under PDR-091b).
* **UPDATING.md**: 1.x → 2.0 migration guide with the old-name → new-name mapping table, marked draft pending PDR-091d.

**Not in this PR** (deferred admin, tracked in the rollout plan): configuring the Trusted Publishers on rubygems.org for both gems — the workflow will fail until that's done, and the bindings must target the post-rename repo name.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -56,6 +56,43 @@ jobs:
           github-token: ${{ secrets.GITHUB_TOKEN }}
           flag-name: ruby-${{ matrix.ruby }}-${{ matrix.protocol }}-${{ matrix.type }}
           parallel: true
+  # Dry run of the release pipeline: the same consistency assertions release.yml's
+  # pre-flight makes, plus a build of both gems — so a version-file or gemspec
+  # regression surfaces on every PR instead of at release time. Publishes nothing.
+  release-dry-run:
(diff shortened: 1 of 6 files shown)


ttypic (approved):
LGTM

PR #455, opened 2026-09-02
Add lockstep release workflow to main (dispatch-only)

umair-ably:
## What

Cherry-picks `.github/workflows/release.yml` (only the workflow file) from #454 onto `main`.

## Why

`workflow_dispatch` workflows are only registered once the file exists on the **default branch**. The Pub/Sub split work merges to `integration/v2` and won't reach `main` until GA, so without this the release workflow cannot be dispatched at all — including the initial run that claims the `ably-pubsub-core` / `ably-pubsub-server` gem names on RubyGems via trusted publishing and proves the OIDC pipeline end to end (`gh workflow run release.yml --ref <branch>`).

## Safety on main

The workflow is inert here:
- Dispatch-only — no push/PR/schedule triggers.
- Its pre-flight requires `core/` and `server/` gem layouts, which don't exist on `main`, so a dispatch at `main` fails before anything is built or pushed.
- Publishing requires RubyGems Trusted Publishers bound to this repo + workflow file, which are not yet configured.

The full release tooling (pre-flight, dry-run CI job, docs) is reviewed in #454; this PR is purely the registration mechanism.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,101 @@
+name: Release
+
+# Releases ably-pubsub-core and ably-pubsub-server in lockstep at the same version
+# (PDR-091b). The pre-flight fails before anything is pushed if the version input,
+# the two gems' version files, or the server->core exact pin disagree.
+#
+# Publishing uses RubyGems trusted publishing (OIDC): both gems must have a Trusted
+# Publisher configured on rubygems.org pointing at this repository and this workflow
+# file. NOTE: the binding is to the repo owner+name, so it must be reconfigured when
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
+name: Release
+
+# Releases ably-pubsub-core and ably-pubsub-server in lockstep at the same version
+# (PDR-091b). The pre-flight fails before anything is pushed if the version input,
```
AndyTWF: Arguably we don't need this PDR reference in the code

AndyTWF (approved):

